PR #35, opened 2024-05-27
Inject tracking service via React context

mpanne:
This refactoring simplifies tracking for shared components:
- shared components use a tracking service loaded from context
- the tracking service implements an interface which defines common / shared methods (like button click tracking)
- each project injects their own implementation of the service into the context
- a default noop implementation is injected which only logs the events on debug

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ Do the following steps to add a new application to this Monorepo.
 2. Depending on the tech stack, make sure that:
    - shared dependencies are configured in the top-level [package.json](./package.json)
    - the `tailwind.config.js` uses the top-level [tailwind.preset.js](./tailwind.preset.js)
-   - the `tsconfig.json` extends the top-level [tsconfig.base.json](./tsconfig.base.json)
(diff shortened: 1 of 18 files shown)


HendrikSchmidt (commented):
Thanks for the proposal, I like the approach! While it does lead to some coupling between app and shared packages I think it is clearly more elegant than what we had before. The only part I am not a big fan of is the use of class syntax, but that's more of a stylistic question and I don't have substantial, objective arguments to support my view :)

And so that we also compare a solution using a Decorator, I think it could look something like this:
```
// TrackedButton.tsx

import Button, {
  type ButtonLinkProps,
  type ButtonProps,
} from "@digitalcheck/shared/components/Button";
import tracking from "services/tracking";

export default function TrackedButton({
  onClickCallback,
  ...props
}: ButtonProps | ButtonLinkProps) {
  const trackedClick = (event: React.MouseEvent<HTMLAnchorElement>) => {
    tracking.trackButtonClick(props.id, props.href);
    onClickCallback && onClickCallback(event);
  };
  return <Button {...props} onClickCallback={trackedClick} />;
}
```

mpanne:
Superseded by PR #36

PR #36, opened 2024-05-28
Migrate from plausible npm package to script

mpanne:
- This is the default for plausible integration
- Remove tracking for button clicks since it is not needed


HendrikSchmidt (approved):
lgtm

PR #51, opened 2024-06-12
Generate pre-filled PDF from user input

monachilada:
This PR enables a PDF to be dynamically generated based on user input.

The basis of this is a new resource route that takes the precheck PDF as a template, and then injects any matching query parameters into the corresponding fields in the PDF before either displaying the PDF or forcing the users browser to download it (also controllable via URL params).

I also started to build the pages and forms required to interact with this functionality, but ran into too many open questions to be able to complete them without additional input from designers. I've left those files intact in order to merge into `main`, but they should be safely tucked away so that they won't interfere with current pages. Work on these can be picked up directly on the main branch after merging this PR.


packages/dito/app/routes/vorpruefung.einschaetzung.$fileName[.pdf].tsx
```diff
@@ -0,0 +1,132 @@
+import { LoaderFunctionArgs } from "@remix-run/node";
+import * as fs from "node:fs";
+import * as path from "node:path";
+import { PDFDocument, PDFForm, PDFTextField } from "pdf-lib";
+
+const FIELD_NAME_POLICY_TITLE: string = "Titel des Regelungsvorhabens";
```
Comment (HendrikSchmidt) [inline packages/dito/app/routes/vorpruefung.einschaetzung.$fileName[.pdf].tsx:6]: The explicit type annotations here are superfluous if I am not mistaken, typescript will infer them automatically: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-annotations-on-variables

packages/dito/app/routes/vorpruefung.einschaetzung.$fileName[.pdf].tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  const positive = ["yes", "unsure"];
+  const negative = "no";
+
+  const pdfData: Uint8Array = await createFilledPDF({
+    title,
+    answers: [
+      {
+        name: FIELD_NAME_PRE_CHECK_POSITIVE_1,
+        checked: positive.indexOf(answers?.["it-system"]) > -1,
```
Comment (HendrikSchmidt) [inline packages/dito/app/routes/vorpruefung.einschaetzung.$fileName[.pdf].tsx:89]: This could be handled more elegantly with `positive.includes(answers?.["it-system"])`, right?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

packages/dito/app/routes/vorpruefung.einschaetzung.tsx
```diff
@@ -0,0 +1,96 @@
...
+            text: assessment.title,
+            look: "ds-heading-02-reg",
+          }}
+          content={{
+            markdown: assessment.subtitle,
+          }}
+        ></Box>
+      </Container>
+      <Container paddingBottom="48">
+        <Form method="post" reloadDocument>
```
HendrikSchmidt, line 48: Would this have worked with a `type=submit` button instead of `reloadDocument`?

packages/dito/app/routes/vorpruefung.einschaetzung.$fileName[.pdf].tsx
```diff
@@ -0,0 +1,132 @@
...
+    download = false,
+    ...answers
+  } = searchParams;
+
+  const positive = ["yes", "unsure"];
+  const negative = "no";
+
+  const pdfData: Uint8Array = await createFilledPDF({
+    title,
+    answers: [
```
Comment (HendrikSchmidt) [inline packages/dito/app/routes/vorpruefung.einschaetzung.$fileName[.pdf].tsx:86]: This could maybe be constructed more implicitly through array indexes but I also don't mind it being an explicit map. Might make sense to pull it out of the function call to clean up the nesting a bit?

packages/dito/app/routes/vorpruefung.einschaetzung.$fileName[.pdf].tsx
Comment (monachilada) [reply packages/dito/app/routes/vorpruefung.einschaetzung.$fileName[.pdf].tsx:6]: True, this was taken from Merlin's code and I didn't question it enough. Good catch 👍

packages/dito/app/routes/vorpruefung.einschaetzung.$fileName[.pdf].tsx
Comment (monachilada) [reply packages/dito/app/routes/vorpruefung.einschaetzung.$fileName[.pdf].tsx:86]: I went for explicit intentionally, so that there's a good overview of which fields we are explicitly filling, since the template might change at some point. Definitely moving the function call makes sense though!

packages/dito/app/routes/vorpruefung.einschaetzung.$fileName[.pdf].tsx
Comment (monachilada) [reply packages/dito/app/routes/vorpruefung.einschaetzung.$fileName[.pdf].tsx:89]: Haha, absolutely. You'll notice this from time to time with my code, I'll often use the very "old school" ways from habit that were better supported back in the day, as I don't have the modern array functions memorised. Great that you call me up on it ;)

packages/dito/app/routes/vorpruefung.einschaetzung.tsx
monachilada: `reloadDocument` is needed specifically to force the browser to download the file... it's a special case I found in the Remix docs somewhere...

PR #78, opened 2024-07-12
Change precheck form handling to work w/o JS

HendrikSchmidt:
Some changes to the way the form submissions are handled so that the pre-check can be completed without JavaScript enabled. I am not a huge fan of this makeshift solution and would probably prefer a move away from `react-hook-form` instead.

Validation isn't supported yet without JS.


monachilada (approved):
I think I'm happy with this if you want to merge it.

HendrikSchmidt:
Superseded by #81

PR #81, opened 2024-07-15
Migrate from React Hook Form to RVF

HendrikSchmidt:
This PR migrates both dito and tool-finder away from react-hook-form and to [rvf](https://www.rvf-js.io/). This package is still in alpha so there will be breaking changes, but it builds upon an existing solution [React Validated Forms](https://www.remix-validated-form.io/) (hence the name) and already feels nicer to work with. Let me know if you feel confident enough to merge this or would rather wait, go with `React Validated Forms v5` or one of the other options below. The author seems to be [reasonably happy with the current state](https://github.com/airjp73/remix-validated-form/discussions/364), but it's still taking him some time it seems.

Other options I considered were:
- [Conform](https://conform.guide/): Looks good, I like the docs, but prefer the server side syntax of rvf. Totally fine, though.
- [Remix Forms](https://remix-forms.seasoned.cc/): Wrapper around `React Hook Form`. Liked the client side syntax a little less.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13220,22 +13220,6 @@
         "react": "^18.3.1"
       }
     },
-    "node_modules/react-hook-form": {
-      "version": "7.52.1",
-      "resolved": "https://registry.npmjs.org/react-hook-form/-/react-hook-form-7.52.1.tgz",
-      "integrity": "sha512-uNKIhaoICJ5KQALYZ4TOaOLElyM+xipord+Ha3crEFhTntdLvWZqVY49Wqd/0GiVCA/f9NjemLeiNPjG7Hpurg==",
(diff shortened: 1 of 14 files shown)


monachilada (approved):
There's a lot to review that I'm not familiar with (tool-finder), but this looks quite good to me. I like the syntax and how compact it is to integrate into the forms themselves. It also feels more stable and less brittle having validation rely on a well established library like Zod. My only suggestion would be creating and ADR to outline the decision process?

HendrikSchmidt:
> My only suggestion would be creating and ADR to outline the decision process?

@monachilada would you be ok with me writing the ADR at a later date and focusing on the implementation this week?

PR #145, opened 2024-09-11
Support page

HendrikSchmidt:


packages/dito/app/routes/download.$fileName.tsx
```diff
@@ -8,6 +8,7 @@ export async function loader({ params, request }: LoaderFunctionArgs) {
   const { fileName } = params;
 
   if (!fileName) {
+    // eslint-disable-next-line @typescript-eslint/only-throw-error
```
monachilada, line 11: I like this. If we're always going to disable this warning, should we just configure it to be ignored at the root level?

HendrikSchmidt: I was also thinking of disabling it globally, but I don't think we can just disable this specific occurrence (`throw new Response`) while still keeping the `only-throw-error` (that I would like to keep). Happy to be proven wrong though!

PR #255, opened 2025-01-20
Refactor result form tests and add missing cases

mpanne:
Looking for feedback!

I used the result form tests as an example to show how our tests could be structured. I applied the following rules: 
- each test case tests one thing only
- use consistent test naming 

The result is a very extensive test class due to the different combinations of possible results of the pre-check. I open this PR for discussion about the following topics: 
- Do you see another way of testing all combinations without adding too much abstraction?
- Do you think testing only one thing per test makes sense for e2e tests? 
- What other rules would you like to see applied for tests?


packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+
+  test("email body contains all answers in positive form", async ({ page }) => {
+    await page
+      .getByLabel("Vorläufiger Arbeitstitel des Vorhabens")
+      .fill("Policy ABCDEFG");
+    const bodyContains = [
+      "In Bezug auf digitale Aspekte führt ihr Regelungsvorhaben zu...",
+      "In Bezug auf Interoperabilität führt ihr Regelungsvorhaben zu...",
+    ];
+    for (let i = 0; i < questions.length; i++) {
```
HendrikSchmidt, line 191: could use a `questions.forEach()` here?

packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+    await expect(page.getByLabel("Arbeitstitel des Vorhabens")).toHaveValue(
+      "Policy 123",
+    );
+    await interceptMailToRedirectAndExpect(page);
+    await page.getByTestId(CREATE_EMAIL_BUTTON).click();
+    await expect(page.getByTestId(EMAIL_INPUT_ERROR)).not.toBeVisible();
+    await expect(page.getByTestId(TITLE_INPUT_ERROR)).not.toBeVisible();
+  });
+
+  test("email subject includes title", async ({ page }) => {
```
HendrikSchmidt, line 127: we could think about separating the following email tests to their own scope and `fill()` in `beforeEach()`

packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+    await interceptMailToRedirectAndExpect(page, undefined, {
+      body: ["Begründung:"],
+    });
+    await page.getByTestId(CREATE_EMAIL_BUTTON).click();
+  });
+});
+
+test.describe("test positive result for digital and negative for interoperability", () => {
+  test.beforeEach(
+    "answer all digital questions with no and all interoperability questions with yes and go to result page",
```
HendrikSchmidt, line 214: description describes the opposite

packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+        "Das Regelungsvorhaben hat einen Digitalbezug und keine Anforderungen der Interoperabilität.",
+      ],
+    });
+    await page.getByTestId(CREATE_EMAIL_BUTTON).click();
+  });
+});
+
+test.describe("test positive result for digital and unsure for interoperability", () => {
+  test.beforeEach(
+    "answer all digital questions with no and all interoperability questions with yes and go to result page",
```
HendrikSchmidt, line 259: description isn't accurate

packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+  test("email body contains all answers for interoperability in unsure form", async ({
+    page,
+  }) => {
+    await page
+      .getByLabel("Vorläufiger Arbeitstitel des Vorhabens")
+      .fill("Policy ABCDEFG");
+    const bodyContains = [
+      "In Bezug auf Interoperabilität ist nicht sicher, ob Ihr Regelungsvorhaben zu Folgendem führt...",
+    ];
+    for (let i = 0; i < questions.length; i++) {
```
HendrikSchmidt, line 309: could use a `questions.filter().forEach()` here

packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+  test("negative reasoning input is visible", async ({ page }) => {
+    await expect(page.getByLabel("Begründung")).toBeVisible();
+  });
+
+  test("error is shown if negative reasoning is empty", async ({ page }) => {
+    await page.getByLabel("Arbeitstitel des Vorhabens").fill("Policy #987");
+    await interceptMailToRedirectAndExpect(page);
+    await page.getByTestId(CREATE_EMAIL_BUTTON).click();
+    await expect(page.getByTestId(NEGATIVE_REASONING_ERROR)).toBeVisible();
+    await expect(page.getByTestId(NEGATIVE_REASONING_ERROR)).toContainText(
```
HendrikSchmidt, line 352: testing for visibility and text seems a bit redundant

packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+    );
+  });
+
+  test("error is shown if negative reasoning is too long", async ({ page }) => {
+    await page.getByLabel("Begründung").fill("A".repeat(501));
+    await page.getByLabel("Arbeitstitel des Vorhabens").fill("Test 123");
+    await interceptMailToRedirectAndExpect(page);
+    await page.getByTestId(CREATE_EMAIL_BUTTON).click();
+    await expect(page.getByTestId(NEGATIVE_REASONING_ERROR)).toBeVisible();
+    await expect(page.getByTestId(NEGATIVE_REASONING_ERROR)).toContainText(
```
HendrikSchmidt, line 363: testing for visibility and text seems a bit redundant

packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+    }
+  });
+}
+
+test.describe("test positive result for digital and interoperability", () => {
+  test.beforeEach(
+    "answer all pre-check questions with yes and go to result page",
+    async ({ page }) => {
+      await page.goto(questions[0].url);
+      for (let i = 0; i < questions.length; i++) {
```
HendrikSchmidt, line 65: could maybe use a `questions.forEach()` here and in the other fixtures

packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+        await page.getByRole("button", { name: "Übernehmen" }).click();
+      }
+    },
+  );
+
+  test("negative reasoning input is not visible", async ({ page }) => {
+    await expect(page.getByLabel("Begründung")).toBeVisible();
+  });
+
+  test("email body contains result title and hint", async ({ page }) => {
```
HendrikSchmidt, line 436: does it contain the hint?

packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+        "Bitte beachten Sie: Wenn Ihr Vorhaben keinen Digitalbezug aufweist, können die Anforderungen der Interoperabilität nicht erfüllt werden",
+      ],
+    });
+    await page.getByTestId(CREATE_EMAIL_BUTTON).click();
+  });
+});
+
+test.describe("test negative result for digital and unsure for interoperability", () => {
+  test.beforeEach(
+    "answer all digital questions with no and all interoperability questions with yes and go to result page",
```
HendrikSchmidt, line 465: not accurate

packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+    await page.getByRole("button", { name: "Übernehmen" }).click();
+    for (let i = 2; i < questions.length; i++) {
+      await page.waitForURL(questions[i].url);
+      await page.getByLabel("Ja").click();
+      await page.getByRole("button", { name: "Übernehmen" }).click();
+    }
+    await page.waitForURL(ROUTE_RESULT.url);
+  });
+
+  test("answers in email body are prefixed by a special character indicating the type of ansewr", async ({
```
HendrikSchmidt, line 513: typo

packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+    page,
+  }) => {
+    await page
+      .getByLabel("Vorläufiger Arbeitstitel des Vorhabens")
+      .fill("Policy ABCDEFG");
+    const bodyContains = [
+      "- " + questions[0].negativeResult,
```
HendrikSchmidt, line 520: just out of curiosity: do you prefer string concatenation over templating and if so, why?

packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+      "+ " + questions[2].positiveResult,
+    ];
+    await interceptMailToRedirectAndExpect(page, {
+      body: bodyContains,
+    });
+    await page.getByTestId(CREATE_EMAIL_BUTTON).click();
+  });
+});
+
+test.describe("test unsure result for digital and positive for interoperability", () => {
```
HendrikSchmidt, line 531: product question: shouldn't this lead to a hint that if operability is concerned, digitalbezug is a given?

packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+      }
+    },
+  );
+
+  test("form is not shown", async ({ page }) => {
+    await expect(page.getByTestId("result-form")).not.toBeVisible();
+  });
+});
+
+test.describe("test unsure result for digital and negative for interoperability", () => {
```
HendrikSchmidt, line 551: Maybe let's check for our support offering here and in the case before and after?

HendrikSchmidt (commented):
Thanks for the initiative and the comprehensiveness of the new tests. Besides the more detail oriented comments (as always, sorry for the brevity, no harshness intended, just tired), these are my larger thoughts:

Things I like:
- no use of custom functions that obfuscate the test code, instead making use of reasonable beforeEach() behaviors
- tests having a single responsibility
- explanatory and consistent test titles

Things that we should discuss
- How could we save more time on test execution
    - by grouping some tests together (that fit semantically) to save some time on repeated navigation with `beforeEach()`
    - going to the result page directly instead of navigating through the precheck entirely all the time? (`beforeAll()` to fill the cookie with answers and then one `goto()` in `beforeEach()`)
- Using [parameterized tests](https://playwright.dev/docs/test-parameterize) to cut down heavily on repeated code for the different combinations

packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+  test("negative reasoning input is not visible", async ({ page }) => {
+    await expect(page.getByLabel("Begründung")).not.toBeVisible();
+  });
+
+  test("error is shown if title is empty", async ({ page }) => {
+    // not filling title
+    await interceptMailToRedirectAndExpect(page);
+    await page.getByTestId(CREATE_EMAIL_BUTTON).click();
+    await expect(page.getByTestId(TITLE_INPUT_ERROR)).toBeVisible();
+    await expect(page.getByTestId(TITLE_INPUT_ERROR)).toContainText(
```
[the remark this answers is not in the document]
mpanne: `toContainText` does not check if the text is visible on page. So especially for error elements (which are hidden most of the time and only shown in case an error occurs) it is important to check both I would say.

packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+      "Bitte geben Sie einen Titel für Ihr Vorhaben an.",
+    );
+  });
+
+  test("error is shown if title is too long", async ({ page }) => {
+    await page.getByLabel("Arbeitstitel des Vorhabens").fill("A".repeat(101));
+    await interceptMailToRedirectAndExpect(page);
+    await page.getByTestId(CREATE_EMAIL_BUTTON).click();
+    await expect(page.getByTestId(TITLE_INPUT_ERROR)).toBeVisible();
+    await expect(page.getByRole("main")).toContainText("kürzeren Titel");
```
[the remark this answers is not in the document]
mpanne: See above

packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+    await expect(page.getByLabel("Arbeitstitel des Vorhabens")).toBeVisible();
+  });
+
+  test("negative reasoning input is not visible", async ({ page }) => {
+    await expect(page.getByLabel("Begründung")).not.toBeVisible();
+  });
+
+  test("error is shown if title is empty", async ({ page }) => {
+    // not filling title
+    await interceptMailToRedirectAndExpect(page);
```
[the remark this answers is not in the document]
mpanne: Because it registers a listener which must happen before the action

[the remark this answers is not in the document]
mpanne: I would love if you have another idea since this violates the "given, when, then" principle. Or "arrange, act, assert", however you wanna call it.

packages/dito/tests/e2e/resultForm.spec.ts
```diff
@@ -0,0 +1,589 @@
...
+    await expect(page.getByTestId(EMAIL_INPUT_ERROR)).not.toBeVisible();
+  });
+
+  test("no error shown when email and title are filled", async ({ page }) => {
+    await page.getByLabel("Ihre E-Mail Adresse").fill("foo@bar.de");
+    await expect(page.getByLabel("Ihre E-Mail Adresse")).toHaveValue(
+      "foo@bar.de",
+    );
+    await page.getByLabel("Arbeitstitel des Vorhabens").fill("Policy 123");
+    await expect(page.getByLabel("Arbeitstitel des Vorhabens")).toHaveValue(
```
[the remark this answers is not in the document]
mpanne: Me neither tbh. This is something I did not touch. I guess the purpose is to check if filling the input fiel did work?

packages/dito/tests/e2e/resultForm.spec.ts
mpanne: Definitely

packages/dito/tests/e2e/resultForm.spec.ts
mpanne: Good catch! Copy & pasta

packages/dito/tests/e2e/resultForm.spec.ts
mpanne: Thank you

packages/dito/tests/e2e/resultForm.spec.ts
mpanne: Thank you!

packages/dito/tests/e2e/resultForm.spec.ts
mpanne: see above

packages/dito/tests/e2e/resultForm.spec.ts
mpanne: I tried but `forEach` expects void and not a Promise (due to the async `await` calls) which leads to an ESLint error.

packages/dito/tests/e2e/resultForm.spec.ts
mpanne: Thank you

packages/dito/tests/e2e/resultForm.spec.ts
mpanne: Thanks

packages/dito/tests/e2e/resultForm.spec.ts
mpanne: Because I'm old and a Java child. 😅  I will try to keep it in mind for the future, I need to get used to it.

packages/dito/tests/e2e/resultForm.spec.ts
mpanne: Yes! But since this is only the test for the result **form**, this is out of scope here. But I actually thought about merging the tests for result form and result page since we have very similar initial conditions for all the test cases (all the result combinations). I didn't do that yet since this refactoring should mainly be a basis for discussion and not a complete refactoring of all tests.

packages/dito/tests/e2e/resultForm.spec.ts
mpanne: Same here. Out of scope since this is not part of the result form.

packages/dito/tests/e2e/resultForm.spec.ts
mpanne: Will cleanup

packages/dito/tests/e2e/resultForm.spec.ts
mpanne: Will be merged in the future as discussed today.

packages/dito/tests/e2e/resultForm.spec.ts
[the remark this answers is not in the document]
HendrikSchmidt: Just for the sake of it, I think the conditional rendering with react should fail `toContainText()`. Anyways, no strong preference, happy either way :)

I would however drop the `getByTestId()` in the second case.

packages/dito/tests/e2e/resultForm.spec.ts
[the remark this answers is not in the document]
HendrikSchmidt: Got it. Maybe this would be something to put in the `beforeAll()` with a comment, then it's a bit clearer that it registers a routing handler?

packages/dito/tests/e2e/resultForm.spec.ts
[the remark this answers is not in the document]
HendrikSchmidt: Possibly, although I expect that this would be a given with just `fill()` already. I'd be in favor of removing the two lines.

packages/dito/tests/e2e/resultForm.spec.ts
HendrikSchmidt: Yes, possibly with nesting `test.describe()` or in their own top level group altogether, both are fine by me

packages/dito/tests/e2e/resultForm.spec.ts
HendrikSchmidt: Didn't think about that! A `for ... of` might work, though? (`map()` probably does, too, but I don't think that's cleaner).

My sonarcube plugin even highlights the code and proposes (not sure if this is the exact same instance of the fixture so use with caution)
```
      for (const element of questions) {
        await page.waitForURL(element.url);
        await page.getByLabel("Ja").click();
        await page.getByRole("button", { name: "Übernehmen" }).click();
      }
```

packages/dito/tests/e2e/resultForm.spec.ts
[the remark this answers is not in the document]
mpanne: The conditional rendering will do the trick but what if someone changes the component behavior to `<InputError id={errorId} className={error ? "hidden" : ""}>` or something like that for some reason. The tests would not fail since the text is in the DOM (just tried locally).
I know it's an edge case but I prefer having an extra LOC for a more bulletproof test. 
Sorry for being such a stubborn here. 🙈

packages/dito/tests/e2e/resultForm.spec.ts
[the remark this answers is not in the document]
mpanne: How would you then declare the expected things? This can only happen in the test method.

packages/dito/tests/e2e/resultForm.spec.ts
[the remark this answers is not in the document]
mpanne: Yes, agree. Will remove

packages/dito/tests/e2e/resultForm.spec.ts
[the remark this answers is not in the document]
mpanne: Right! Agree.

packages/dito/tests/e2e/resultForm.spec.ts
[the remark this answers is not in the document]
HendrikSchmidt: Totally overlooked that, you're right. Maybe just give it a name that makes it clearer that it registers a routing handler for the mail then and leave it as is? Don't have a better idea. Something like `registerMailHandlerAndExpect()`?

PR #296, opened 2025-02-26
Remove tool finder

HendrikSchmidt:


README.md
```diff
@@ -1,24 +1,29 @@
...
+We aim to use the current active [LTS version of nodejs](https://nodejs.dev/en/about/releases/).
 There is a `.node-version` file to simplify setup using [nodenv](https://github.com/nodenv/nodenv).
 
 ### Dependencies
 
 Install the dependencies using npm.
 
 ```bash
-npm install
+npm install --ignore-scripts @strapi/blocks-react-renderer
```
lsolcher, line 26: Why is this added?

doc/adr/0009-use-monorepo.md
```diff
@@ -1,10 +1,10 @@
-# 3. Use a Monorepo for All Digitalcheck Applications
+# 9. Use a Monorepo for All Digitalcheck Applications
 
-Date: 2024-05-23
+Date: 2024-05-24
```
mpanne, line 3: Why did the dates change in some ADRs (not only this one)?

Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+COPY package.json package-lock.json start.sh ./
+COPY --from=production-dependencies /src/node_modules/ ./node_modules/
+COPY --from=build /src/build/ ./build/
+# We need to explicitly bring in the public folder here, so that dynamic PDF generation can happen on the server 
+COPY --from=build /src/public/ ./public/
+
+# Make the start script executable
+RUN chmod +x ./start.sh && \
+    # Ensure the node user owns all files in the working directory
+    chown -R node:node /home/node/src
```
mpanne, line 49: Remove indent

package.json
```diff
@@ -1,43 +1,52 @@
 {
-  "name": "digitalcheck-apps",
```
mpanne, line 2: Keep until renaming (will be renamed when renaming the repo I guess?) or change to "digitalcheck-dito" as discussed.

package.json
```diff
@@ -1,43 +1,52 @@
...
+  "name": "digitalcheck",
   "private": true,
-  "homepage": "https://digitalservicebund.github.io/digitalcheck-apps/",
+  "homepage": "https://digitalservicebund.github.io/digitalcheck/",
```
mpanne, line 4: This does not exist. How about "https://erarbeiten.digitalcheck.bund.de/"?

README.md
```diff
@@ -101,41 +123,29 @@ npm run build
...
-     - `test:a11y`: run a11y tests
-     - `tests`: run all tests
-   - a `Dockerfile` which builds a Docker container from your application (and if needed a `Dockerfile.dockerignore`)
-     - also add your service to the [docker-compose.yml](./docker-compose.yml) file
-4. Add the application code. It should now be able to import shared components.
-5. Before proceeding with the next step, make sure the infrastructure of the application is in place.
-   Follow the steps in the [digitalcheck-apps-infra README.md](https://github.com/digitalservicebund/digitalcheck-apps-infra).
-6. Update the [pipeline.yml](./.github/workflows/pipeline.yml) by copying an existing filter in the job `changes` and substituting the new application name.
-7. You are done! The new application will be build and deployed on commit.
+The pipeline includes a job to build the Docker Image and push it to GitHub Packages. This job is currently deactivated. To enable it you need to remove the `&& false` from the end of the `if` predicate of the `build-and-push-image` job.
```
mpanne, line 139: What? Why is it deactivated or why is the comment here?

mpanne (changes requested):
Great work, thank you! 
Some cleanup tasks todo but nothing groundbreaking.

README.md
HendrikSchmidt: During installation `@strapi/blocks-react-renderer` runs a `husky` (some kind of code quality tool) script that fails if you don't have `husky` installed. With this flag you can skip that part.

.github/workflows/nightly-tests.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
```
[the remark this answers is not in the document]
HendrikSchmidt: Removed

.github/workflows/check-and-test.yml
```diff
@@ -3,10 +3,6 @@ name: Check and Test
 on:
   workflow_call:
     inputs:
-      package:
-        required: true
-        type: string
-        description: "The Package to check and test."
       shouldRunIntegrationTests:
```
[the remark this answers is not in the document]
HendrikSchmidt: Removed

README.md
HendrikSchmidt: I copied this part without checking whether this still holds true and forgot to do the check. Doesn't seem to be relevant anymore, I'll remove it.

package.json
HendrikSchmidt: Changed to your proposition.

doc/adr/0009-use-monorepo.md
HendrikSchmidt: Ok, hear me out: First, I wanted to consolidate the ADRs by merging them and sorting by date. But then I noticed this led to the wrong order of the ADRs: Decisions that were taken later would come earlier than others. So instead of just reordering the decisions I also changed the date to at least reflect the correct order the decisions were taken in – but didn't spend time on researching the correct date nor doing any other checks like which decisions were actually taken on the same day. As an example, ADR 16 has to come before 17, as 17 discusses GraphQL tooling. It was, however, written a whole month earlier, so I changed the date to make a bit more sense (I noticed this only after your review).

Dockerfile
HendrikSchmidt: Why? Don't you think it looks better like this? :D

mpanne: Ahh, now I get it. Looked like a mistake at first glance. Yes, leave it.

doc/adr/0009-use-monorepo.md
mpanne: Ok, I get your intention. Thing is, that the dates in the ADRs do not reflect the date of the decision being taken but the date of the ADR being written. If you now change them, they don't reflect anything. 😄 I don't really care, the exact date does not really matter. Maybe we can to use the date of the decision being taken in the future. So whenever we take a decision, we either create a ticket to write the ADR and write down the date or we create an empty ADR file with a date and the rest ist TDB.

mpanne (approved):

doc/adr/0009-use-monorepo.md
HendrikSchmidt: Totally d'accord with creating a stub each time we take a decision, let's do that!

PR #320, opened 2025-03-20
Improve content management with TOML or YAML

mpanne:
**Pain:** 
Working with the content.ts is not fun anymore. The file now has about 2400 LoC. 

**Goal:** 
Use a different representation of the content in a more structured and human readable format.

**Solution Approach:** 
- split `content.ts` into different files named by the route they contain content for
- use TOML or YAML instead of TypeScript objects to represent the textual content
- use vite plugins to be able to import TOML and YAML and automatically transform the content to a TypeScript object
- write a custom vite plugin to automatically generate `.d.ts` files with interfaces from YAML content to ensure type safety
- add an `Icon` component that transforms the name of the icon to be used as string into a React component


mpanne:
**Decision:** We do not use YAML or TOML as content representation format. We do split the `content.ts` file into route specific TypeScript files. 

**Reasoning:**
- TypeScript does not natively support those and types must dynamically be generated every time the content changes which kills the DX
- the benefit does not outweigh the drawbacks of supporting YAML or TOML

PR #327, opened 2025-03-25
Upgrade and refactor lint 9

lsolcher:
Upgrades eslint to v9 and transforms the config to a flat config


app/components/Button.tsx
```diff
@@ -23,6 +23,7 @@ export interface ButtonLinkProps
 
 function formatIcon(icon: ReactElement | undefined) {
   if (!icon) return undefined;
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
mpanne, line 26: How about using a specific type (icon: ReactElement<SVGProps<SVGSVGElement>>) instead?

app/components/ParagraphList.tsx
```diff
@@ -35,6 +35,7 @@ const extractTextParts = (children: ReactNode) => {
   if (!children || typeof children !== "object" || !("props" in children)) {
     return null;
   }
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
mpanne, line 38: Can we use a specific type here as well?

eslint.config.mjs
```diff
@@ -0,0 +1,159 @@
...
+          ],
+        },
+      ],
+    },
+  },
+  // Additional Rules for test files
+  {
+    files: [
+      "**/a11y/*.{spec,test}.{js,ts,jsx,tsx}",
+      "**/e2e/*.{spec,test}.{js,ts,jsx,tsx}",
```
mpanne, line 127: Why do we still include .js and .jsx here?

eslint.config.mjs
```diff
@@ -0,0 +1,159 @@
...
+    ],
+    extends: [playwright.configs["flat/recommended"]],
+    plugins: { "testing-library": testingLibrary },
```
mpanne, line 130: can we also extend here instead of using plugin?

mpanne (approved):
Legit, thank you!

eslint.config.mjs
lsolcher: That gives us ✖ 136 problems (136 errors, 0 warnings) 😬 and since we already have quite a few rules defined for this plugin, i thought there was a good reasoning behind it from the original creator

eslint.config.mjs
lsolcher: right, no need for this anymore!

app/components/ParagraphList.tsx
lsolcher: Talked to Hendrik about this one, here is it not really feasible to set a type, leads to other issues so ignoring the warning here seems the best approach

PR #357, opened 2025-04-17
Implementation of header menu

lsolcher:
Implementation of the Dropdown header menu. 

- PageHeader: Renders header, incorporating navigation menus for desktop and mobile, while managing the open/closed states of navigation elements.
- Dropdown: Creates an interactive dropdown menu component that displays a list of items when triggered, handling different visual variants for desktop and mobile layouts and managing its own expanded state.
- DropdownContentList: Renders the interactive links within a dropdown menu by mapping data to DropdownItem components and structuring them.
- DropdownItem: Displays an individual clickable item within a dropdown menu.
- Design + Behaviour: https://www.figma.com/design/HiiKWvM5xBlNCM9wXYvInb/%F0%9F%9F%A6-Digitalcheck-Komponenten-Library?node-id=91-620&p=f&t=LS2yWmbZi9QPYH3W-0


app/root.tsx
HendrikSchmidt, line 1: Could be the right time to fix our messy rendering here and move to a `Layout` export, but I am also fine for doing this in a separate step, just wanted to raise the issue again.

app/components/PrinciplesDisplay.tsx
HendrikSchmidt, line 1: I have a similar feeling with this as with `MethodStepsItems`:
> I'm not a big fan of how this component mixes display logic with what content should be displayed but I am having trouble expressing what exactly my issue is. It just feels like it adds an extra layer of complexity that might not be needed, maybe we can talk about that.

app/components/DrodownItem.tsx
HendrikSchmidt, line 1: Couldn't an implementation with `NavLink` still make sense and work for principle pages as well? If we link to the parent page (`/prinzipien`) and forward from there to the first principle the link should stay active if I get the documentation right.

app/components/Dropdown.tsx
HendrikSchmidt, line 1: I think the naming of the Dropdown components might be a bit confusing as they can only be used for the menu, wdyt?

app/components/DropdownContentList.tsx
HendrikSchmidt, line 1: Why do we need this split into lists and none lists? Could it also be `ol`s and `ul`s?

HendrikSchmidt (changes requested):
The menu is looking great already, but I have a few concerns before I go into a more detailed review:
- The menu is not working for me with Adblock enabled, probably an issue with the way that the tracking is handled?
- I am getting a 500 for the support links
- I think we should change the components folder strucure: `MethodStepItems` and `PageHeader` don't fit into well into the concept of reusable components and the folder is growing to be confusing, but there are other offenders and this might be something for a latter PR?
- What is the point of the new "Grundlagen" method page in contrast to the already existing one? (probably more of a design question)
- Same goes for the second principles page

Also, as a general note (I am also a offender here) maybe we could try to keep our PRs smaller/more focused as it gets hard to review this amount of lines and I sometimes lose track of the things I want to look at. This could have maybe been split into the menu and then adding the new fundamentals pages?

app/components/Dropdown.tsx
```diff
@@ -0,0 +1,126 @@
...
+      {isExpanded && (
+        <div
+          className={panelClasses}
+          id={`${elementId}-content`}
+          role={!isMobile ? "listbox" : "region"}
+          aria-labelledby={`${elementId}-button`}
+        >
+          {hasSupport && <DropdownSupportItem mobile={isMobile} />}
+          {/*
+          // TODO: add second row for more than 4 items
```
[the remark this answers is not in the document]
lsolcher: No, this will go in another iteration since we don't have items with more than 3 items yet. I have created a tech debt ticket for this

app/routes/grundlagen.tsx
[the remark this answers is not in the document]
lsolcher: It exists to prevent 404 when clicking on the `grundlagen` breadcrumb. We plan to create the grundlagen page and add further content to grundlagen in further iterations. For the moment we agreed it should just redirect. Renamed the file

app/routes/grundlagen_.fuenf-prinzipien.tsx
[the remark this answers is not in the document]
lsolcher: The concept is that we have one method and principle site which is included in the erarbeiten workflow (the current ones). The new pages are part of the new grundlagen content, which are there to give people an overview of all methods and also the principles. They are for two different use cases. Renamed the files

lsolcher:
The issue with the adblock is interesting. I checked and it works on staging with adblock enabled. Investigating

lsolcher:
Ok the issue with the dropdown menu not working with adblock on localhost because we had three scripts blocked: plausible (i changed that with the last push), Pageview.js and Fingerprint.js which come from `@digitalservicebund/icons`. I tested to run it without these two icons included and it works. I believe because the names follow a pattern adBlock would block and the packages get bundled for prod build so the names don't exist anymore on the environment. Since this only happens on localhost with adblock I don't see it as a blocking issue and would rather whitelist localhost but let's talk about it

app/components/PrinciplesDisplay.tsx
lsolcher: Ticket created

app/components/DrodownItem.tsx
lsolcher: Done with the caveat that we need to check the urls with # manually, since [this is not supported by NavLink](https://github.com/remix-run/react-router/issues/10325):
`  const mapDataToItems = (option: DropdownItemProps, index: number) => {
    const renderContent = ({ isActive }: { isActive: boolean }) => {
      const browserHasHash = location.hash !== "";
      // Don't mark support subsection as active items
      const finalIsActive = isActive && !browserHasHash;

      return (
        <DropdownItem
          number={isList ? index + 1 : undefined}
          newContent={option.newContent}
          title={option.title}
          content={option.content}
          isNewTitle={option.isNewTitle}
          isActive={finalIsActive}
          className={option.className}
        />
      );
    };

    return (
      <li key={index}>
        <NavLink
          to={option.href!}
          role="option"
          onClick={onItemClick}
          aria-label={option.title}
        >
          {renderContent}
        </NavLink>
      </li>
    );
  };`

app/components/DropdownSupportItem.tsx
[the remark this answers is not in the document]
lsolcher: Ticket created

app/routes/unterstuetzung/route.tsx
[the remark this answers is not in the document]
lsolcher: Done

tests/e2e/general.spec.ts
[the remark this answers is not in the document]
lsolcher: changed to:
```
    // Remove landing page and grundlagen page
    if (
      route.url.endsWith(".pdf") ||
      route.url === "grundlagen" ||
      route.url === "/"
    ) {
      return;
    }
```

lsolcher:
[The latest changes](https://github.com/digitalservicebund/digitalcheck-dito/compare/de6398fd26b18cac02cbfa1d6def5e22337b5d17...96f6c9a8dce84f39ea1a3aa69cb09b5cdc6a0be3):
- Use NavLink and NavLink active logic for DropdownItems. Add manual logic to not mark urls with # active
- Get rid of allHrefs for checking if a parent dropdown menu item is active. Add logic to check the base path
- Rename Dropdown components
- Fix scrolling logic
- Add prinzipien.tsx page and rename grundlagen routes
- Unify list logic for DropdownItems
- Unify testing logic for breadcrumbs test
- Add readme for local development with adblocker

tests/e2e/general.spec.ts
```diff
@@ -9,9 +9,13 @@ test.describe("test breadcrumbs and titles", () => {
...
   });
 
-  // Remove landing page
-  allRoutes.slice(1).forEach((route) => {
-    if (route.url.endsWith(".pdf")) {
+  allRoutes.forEach((route) => {
+    // Remove landing page and grundlagen page
+    if (
+      route.url.endsWith(".pdf") ||
+      route.url === "grundlagen" ||
```
HendrikSchmidt, line 16: Why are we excluding grundlagen here again? They also have breadcrumbs, right?

app/root.tsx
```diff
@@ -177,10 +174,14 @@ export function ScrollAndFocus() {
   const mainRef = useRef<HTMLDivElement>(null);
 
   useEffect(() => {
-    window.scrollTo({ top: 0, behavior: "instant" });
-    requestAnimationFrame(() => {
-      mainRef.current?.focus();
-    });
+    if (window.location.hash && !window.location.hash.includes("#")) {
```
HendrikSchmidt, line 177: Could you add one or two comments here what happens? I can't completely follow :)

Also, how can `window.location.hash && !window.location.hash.includes("#")` be true at the same time?

app/resources/content/grundlage-methoden.ts
```diff
@@ -0,0 +1,25 @@
...
+        content:
+          "Damit es möglichst einfach für Sie ist, Digitaltauglichkeit in Ihrer Regelung herzustellen, schlagen wir Ihnen einen Methoden-Fahrplan vor. Dieser ist in sich sinnvoll aufgebaut.\n" +
+          "Einige der Methoden sind thematische Erweiterungen der Arbeitsschritte, die Sie bereits einsetzen. Ein Beispiel dafür ist das Einbeziehen von umsetzenden Akteuren.",
+      },
+      {
+        headline: {
+          text: "Ein Beispiel",
+        },
+        content:
+          "Das hier ist ein Fahrplan. Sie können sich so eine Vorstellung davon machen, wie er aufgebaut ist. In der Arbeit damit bekommen Sie zu jedem Punkt Methoden-Anleitungen, Vorlagen und Tipps für die Arbeit damit.",
```
HendrikSchmidt, line 21: `Das hier ist ein Fahrplan.` klingt ein wenig ulkig :D

app/components/PageHeader.tsx
```diff
@@ -0,0 +1,226 @@
...
+}
+
+interface HeaderItem {
+  text: string;
+  overlayContent: SubItem[];
+  hasSupport?: boolean;
+  isList?: boolean;
+}
+
+const findActiveParentItemText = (
```
HendrikSchmidt, line 27: This function could be made simpler by using `find()` and maybe inlining it completely like so:
```
      const parentItemText = header.items.find((item) =>
        isParentItemActive(item, location.pathname),
      )?.text;
      setActiveDropdownId(parentItemText ?? null);
```

app/components/PageHeader.tsx
```diff
@@ -0,0 +1,226 @@
...
+const isParentItemActive = (item: HeaderItem, path: string): boolean => {
+  return item.overlayContent.some((subItem) => {
+    const normalizedItemPath = normalizePathname(subItem.href);
+    const normalizedCurrentPath = normalizePathname(path);
+    // Exact match
+    if (normalizedCurrentPath === normalizedItemPath) {
+      return true;
+    }
+    // Prefix match (current path starts with item href + '/')
+    return normalizedCurrentPath.startsWith(normalizedItemPath + "/");
```
HendrikSchmidt, line 49: Shouldn't the prefix match be enough in any case? It doesn't really matter if we have an exact match, right?

app/components/PageHeader.tsx
```diff
@@ -0,0 +1,226 @@
...
+    };
+  }, []);
+
+  // Toggle dropdown state
+  const toggleDropdown = (itemText: string) => {
+    setActiveDropdownId((current) => (current === itemText ? null : itemText));
+  };
+
+  // Close all open dropdowns
+  const closeAll = () => {
```
HendrikSchmidt, line 95: There can only ever be one open dropdown, right? So aren't we only effectively closing one dropdown?

HendrikSchmidt: Ahh or does it mean closing both dropdown and menu?

app/components/PageHeader.tsx
```diff
@@ -0,0 +1,226 @@
...
+    setActiveDropdownId(null);
+    setMobileMenuOpen(false);
+  };
+
+  // Opens and closes mobile menu
+  const toggleMobileMenu = () => {
+    const isOpen = !mobileMenuOpen;
```
HendrikSchmidt, line 102: I find `isOpen` a bit confusing here? Maybe something along these lines: 
```
    if (!mobileMenuOpen) {
      setMobileMenuOpen(true);
      const parentItemText = header.items.find((item) =>
        isParentItemActive(item, location.pathname),
      )?.text;
      setActiveDropdownId(parentItemText ?? null);
    } else {
      closeAll();
    }
```

app/components/PageHeader.tsx
```diff
@@ -0,0 +1,226 @@
...
+    <>
+      {showOverlay && (
+        <div
+          className="fixed inset-0 z-20 bg-[#282828] opacity-63"
+          onClick={closeAll}
+          aria-hidden="true"
+        />
+      )}
+      <header className="relative" ref={headerRef}>
+        <div className="relative">
```
HendrikSchmidt, line 127: Is this `div` necessary?

app/components/PageHeader.tsx
```diff
@@ -0,0 +1,226 @@
...
+          <div className="relative z-30 flex min-h-[72px] items-stretch justify-between bg-white pl-16 lg:px-16">
```
HendrikSchmidt, line 128: Could `min-` and `items-stretch` be removed?

app/components/PageHeader.tsx
```diff
@@ -0,0 +1,226 @@
...
+            {/* Logo and title */}
+            <div className="flex items-center space-x-8">
+              <Link to={ROUTE_LANDING.url}>
+                <img
+                  src="/logo/bund-logo.png"
+                  alt="Logo des Bundes"
+                  width={54}
+                />
+              </Link>
+              <div className="ds-stack-0 ml-16 flex flex-col -space-y-4 max-lg:hidden">
```
HendrikSchmidt, line 138: What is `ds-stack-0` doing here and could the `div` and `p` be merged?

HendrikSchmidt: `<p className="ds-label-01-bold ml-16 flex flex-col -space-y-4 text-xl max-lg:hidden">`

tests/e2e/general.spec.ts
lsolcher: True, this is a leftover, removed!

app/components/PageHeader.tsx
lsolcher: Nice refactor

app/components/PageHeader.tsx
lsolcher: Yep :)

app/components/PageHeader.tsx
lsolcher: Another leftover for when we had two lines of text in the header, same as the -space-y-4. Removed it!

app/components/PageHeader.tsx
```diff
@@ -0,0 +1,226 @@
...
+                aria-expanded={mobileMenuOpen}
+                aria-controls="mobile-menu"
+                tabIndex={0}
+              >
+                {mobileMenuOpen ? <MenuOpen /> : <MenuOutlined />}
+              </button>
+            </div>
+          </div>
+          {/* Mobile Navigation */}
+          <div
```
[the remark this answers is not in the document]
lsolcher: It should be, changed

app/components/PageHeader.tsx
```diff
@@ -0,0 +1,226 @@
...
+            <div className="flex items-center space-x-16 lg:hidden">
+              <a
+                className="border-b-[4px] border-transparent"
+                href={`tel:${header.contact.number}`}
+              >
+                <PhoneOutlined />
+              </a>
+              <button
+                className={twMerge(
+                  "h-full cursor-pointer border-b-[4px] border-transparent px-16 hover:bg-blue-100 focus:border-b-[4px] focus:bg-blue-100",
```
HendrikSchmidt: `cursor-pointer` seems unnecessary? also `border-b-[4px]` is repeated three times

app/components/PageHeader.tsx
```diff
@@ -0,0 +1,226 @@
...
+                tabIndex={0}
+              >
+                {mobileMenuOpen ? <MenuOpen /> : <MenuOutlined />}
+              </button>
+            </div>
+          </div>
+          {/* Mobile Navigation */}
+          <div
+            id="mobile-menu"
+            className={`absolute right-0 left-0 z-40 rounded-b-md border-t-1 border-gray-600 bg-white drop-shadow-[4px_4px_12px_rgba(0,0,0,0.06)] ${
```
HendrikSchmidt: could you use `twJoin()` here instead of a template string?

app/components/PageHeader.tsx
```diff
@@ -0,0 +1,226 @@
...
+                <p className="ds-stack-0 ds-label-01-bold text-xl">
+                  {header.title}
+                </p>
+              </div>
+            </div>
+
+            {/* Desktop Navigation */}
+            <nav className="flex items-center max-lg:hidden">
+              {header.items.map((item) => {
+                return (
```
HendrikSchmidt: you can return without a `return` statement here :)

app/components/PageHeader.tsx
```diff
@@ -0,0 +1,226 @@
...
+          {/* Mobile Navigation */}
+          <div
+            id="mobile-menu"
+            className={`absolute right-0 left-0 z-40 rounded-b-md border-t-1 border-gray-600 bg-white drop-shadow-[4px_4px_12px_rgba(0,0,0,0.06)] ${
+              mobileMenuOpen ? "overflow-y-auto" : "invisible"
+            }`}
+            aria-hidden={!mobileMenuOpen}
+          >
+            {header.items.map((item) => {
+              return (
```
HendrikSchmidt, line 128: see above :)

app/components/PageHeader.tsx
```diff
@@ -0,0 +1,226 @@
...
+                <DropdownMenu
```
HendrikSchmidt, line 129: Quite a bit of repetition with the component above, but also not sure whether a wrapper would be nicer here.

tests/e2e/general.spec.ts
```diff
@@ -9,9 +9,9 @@ test.describe("test breadcrumbs and titles", () => {
     await expect(page.getByTestId("breadcrumbs-menu")).toBeHidden();
   });
 
-  // Remove landing page
-  allRoutes.slice(1).forEach((route) => {
-    if (route.url.endsWith(".pdf")) {
+  allRoutes.forEach((route) => {
+    // Remove landing page and grundlagen page
```
HendrikSchmidt, line 13: comment seems inaccurate?

app/components/DropdownMenu.tsx
```diff
@@ -0,0 +1,121 @@
...
+  };
+
+  const handleMouseLeave = () => {
+    if (!isMobile && isExpanded) {
+      onToggle();
+    }
+  };
+
+  return (
+    <div
```
HendrikSchmidt, line 87: Sonar complains here:
```
Avoid non-native interactive elements. If using native HTML is not possible, add an appropriate role and support for tabbing, mouse, keyboard, and touch inputs to an interactive content element.
```

app/components/DropdownMenu.tsx
```diff
@@ -0,0 +1,121 @@
...
+        className={buttonClasses}
+        id={`${elementId}-button`}
+        aria-haspopup="true"
+        aria-expanded={isExpanded}
+        aria-controls={`${elementId}-inhalt`}
+      >
+        {label} {isExpanded ? <ExpandLessOutlined /> : <ExpandMoreOutlined />}
+      </button>
+      {isExpanded && (
+        <div
```
HendrikSchmidt, line 104: See above. Also, have you explore a [popover](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/popover) for this?

app/components/DropdownContentList.tsx
```diff
@@ -0,0 +1,65 @@
...
+  data,
+  isList = false,
+  onItemClick,
+  isMobile = false,
+}: Readonly<DropdownContentListProps>) {
+  const location = useLocation();
+  const mobileElementStyle = "py-8 pr-8 pl-16 border-b-[1px] border-gray-600 ";
+  const listClasses = twMerge(
+    "list-unstyled list-none",
+    isMobile ? mobileElementStyle : "",
```
HendrikSchmidt, line 24: `&&` would suffice here :)

app/components/DropdownContentList.tsx
```diff
@@ -0,0 +1,65 @@
...
+          className={option.className}
+        />
+      );
+    };
+
+    return (
+      <li key={index}>
+        <NavLink
+          to={option.href!}
+          role="option"
```
HendrikSchmidt, line 50: is this role correct for a nav menu?

app/components/DropdownContentList.tsx
```diff
@@ -0,0 +1,65 @@
...
+export default function DropdownContentList({
+  data,
+  isList = false,
+  onItemClick,
+  isMobile = false,
+}: Readonly<DropdownContentListProps>) {
+  const location = useLocation();
+  const mobileElementStyle = "py-8 pr-8 pl-16 border-b-[1px] border-gray-600 ";
+  const listClasses = twMerge(
+    "list-unstyled list-none",
```
HendrikSchmidt, line 23: isn't `list-unstyled` enough here?

app/components/DropdownContentList.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+    return (
+      <li key={index}>
+        <NavLink
+          to={option.href!}
+          role="option"
+          onClick={onItemClick}
+          aria-label={option.title}
+        >
+          {renderContent}
```
HendrikSchmidt, line 54: I think I would prefer inlining here instead of a single use function, but I am gladly convinced otherwise

app/components/PageHeader.tsx
lsolcher: Removed the unneeded borders but I think the cursor-pointer is still a good thing here, thinking about smaller browser windows (like half window)

app/components/PageHeader.tsx
lsolcher: yes

app/components/PageHeader.tsx
lsolcher: done

app/components/PageHeader.tsx
lsolcher: done

tests/e2e/general.spec.ts
lsolcher: true, fixed

app/components/DropdownMenu.tsx
```diff
@@ -0,0 +1,121 @@
...
+  onItemClick,
+  isActiveParent,
+}: Readonly<DropdownProps>) {
+  const isMobile = variant === "mobile";
+  const elementId = `dropdown-${label}`;
+
+  // Transparent borders to avoid layout shifts
+  const buttonClasses = twMerge(
+    "flex cursor-pointer items-center hover:bg-blue-100",
+    !isMobile &&
```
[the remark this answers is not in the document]
lsolcher: true, that's cleaner

app/components/DropdownMenu.tsx
```diff
@@ -0,0 +1,121 @@
...
+    : "";
+
+  const containerClasses = twMerge(
+    "relative",
+    isMobile ? "w-full" : "h-full max-lg:hidden",
+    className,
+  );
+
+  useEffect(() => {
+    const handleKeyDown = (event: KeyboardEvent) => {
```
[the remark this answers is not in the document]
lsolcher: Yes, that's not needed here, removed it

app/components/DropdownContentList.tsx
```diff
@@ -0,0 +1,65 @@
+import { NavLink, useLocation } from "react-router";
+import DropdownItem, {
+  type DropdownItemProps,
+} from "~/components/DrodownMenuItem.tsx";
+import twMerge from "~/utils/tailwindMerge";
+
+export type DropdownContentListProps = {
+  data: DropdownItemProps[];
+  isList?: boolean;
```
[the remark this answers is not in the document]
lsolcher: Renamed

app/components/DropdownContentList.tsx
lsolcher: Done

app/components/PageHeader.tsx
lsolcher: Yes, renamed for clarity

app/components/DropdownContentList.tsx
lsolcher: it is!

app/root.tsx
lsolcher: You're right, that didn't make sense, didn't think right. Fixed that

app/components/PageHeader.tsx
lsolcher: Yes, the first one was not needed, removed

app/components/PageHeader.tsx
lsolcher: changed that

app/routes/beispiele.prinzipien._index.tsx
```diff
@@ -0,0 +1,9 @@
+import { redirect } from "react-router";
+import allRoutes from "~/resources/allRoutes.ts";
+import { ROUTE_PRINCIPLES } from "~/resources/staticRoutes.ts";
+
+export function loader() {
+  return redirect(
+    allRoutes.filter((route) => route.parent === ROUTE_PRINCIPLES.url)[0].url,
```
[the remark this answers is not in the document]
lsolcher: Changed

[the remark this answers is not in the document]
lsolcher: Ok this breaks the breadcrumbs test since we have duplicate route defined from allRoutes for the first principle. I'll create a ticket for me and change this once the breadcrumbs are merged, then this won't be an issue any more

app/components/PageHeader.tsx
lsolcher: Since the variant is basically the difference here, it makes sense for me to extract it to a renderFn, I did that so we are DRY

app/components/DropdownContentList.tsx
lsolcher: Right, changed it to menuitem and the parent to menu

app/components/DrodownMenuItem.tsx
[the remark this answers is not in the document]
lsolcher: True it is only being used once currently but this might change with the implementation of TabNavigation. I'm not yet sure if and which components we can re-use but I would wait until that's clear with refactoring this if you're fine with that?

PR #363, opened 2025-04-22
Change breadcrumbs handling

HendrikSchmidt:
- Introduce new breadcrumb handling: Use a handle where it exists (e.g. dynamic routes like questions or examples), use the statically defined routes otherwise. Also works with trailing slashes and SSG now.
- Make route structure more consistent with `parent` + `parent._index` everywhere (necessary for correct breadcrumb handling too)
- Remove `allRoutes` and simplify to `staticRoutes` only
- Introduce a new helper function for route creation that streamlines URL building and parent definitions
- Improve naming of some select routes
- Add eslint rules for all allowed react-router exports


app/resources/staticRoutes.ts
```diff
@@ -1,166 +1,178 @@
...
   ROUTE_LANDING,
+  ROUTE_SUPPORT,
+  ROUTE_IMPRINT,
+  ROUTE_PRIVACY,
+  ROUTE_A11Y,
+  ROUTE_SITEMAP,
   ROUTE_PRECHECK,
-  ROUTE_GENERAL_INFO,
+  ROUTE_PRECHECK_INFO,
   ROUTE_RESULT,
```
HendrikSchmidt, line 161: change to PRECHECK_RESULT

app/routes/beispiele.regelungen.$regelung.tsx
```diff
@@ -34,6 +35,13 @@ export const meta = ({ matches }: MetaArgs) => {
   return prependMetaTitle(ROUTE_REGELUNGEN.title, matches);
 };
 
+export const handle = {
+  breadcrumb: (match: UIMatch) => ({
+    title: "Regelung",
```
HendrikSchmidt, line 40: change to regelungsbeispiel?

app/root.tsx
```diff
@@ -267,7 +266,7 @@ const PageHeader = ({
     </div>
     {includeBreadcrumbs && (
       <Background backgroundColor="blue">
-        <Breadcrumbs breadcrumbs={routes} useIconForHome />
```
HendrikSchmidt, line 270: put background in breadcrumbs

app/routes/vorpruefung.$questionId/route.tsx
```diff
@@ -26,6 +31,11 @@ export const meta = ({ matches }: MetaArgs) => {
   return prependMetaTitle(ROUTE_PRECHECK.title, matches);
 };
 
+export const handle = {
```
HendrikSchmidt, line 34: replace with routes

app/components/Breadcrumbs.tsx
```diff
@@ -1,88 +1,60 @@
...
-                  {displayElement}
-                </Link>
-              )}
-            </span>
-          </div>
-        );
-      })}
+      <Link
+        to={ROUTE_LANDING.url}
+        className="ds-link-01-bold fill-blue-800 outline-blue-800"
```
lsolcher, line 39: is ds-link-01-bold needed here, since the text is is visible for sr only?

app/resources/content/beispiele-regelungen.ts
```diff
@@ -1,4 +1,5 @@
 export const examplesRegelungen = {
+  breadcrumb: "Regelungsbeispiel",
```
lsolcher, line 2: Just as a general consideration maybe it's better to have the Regelungsbeispiele under Prinzipien, so you would have `Beispiele/Prinzipien/Regelungsbeispiel`, since the user probably came from the Prinzipien page when landing here. What do you think?

tests/a11y/general.spec.ts
```diff
@@ -1,37 +1,37 @@
...
+  ROUTES.filter(
+    (route) =>
+      !route.url.endsWith(".pdf") && !route.url.endsWith(ROUTE_SUPPORT.url),
+  ).forEach((route) => {
+    test(`check a11y of ${route.title}`, async ({ page }) => {
+      // Listen for redirects and update URL if needed
+      const response = await page.goto(route.url);
+
+      if (
+        (response !== null && response.status() === 302) ||
```
lsolcher, line 20: can be response !== null && (response.status() === 301 || response.status() === 302)

tests/snapshots/general.spec.ts
```diff
@@ -1,34 +1,32 @@
...
 
-        if (
-          (response !== null && response.status() === 302) ||
-          (response !== null && response.status() === 301)
-        ) {
-          const redirectedUrl = response.headers()["location"];
-          await page.goto(redirectedUrl);
-        }
+      if (
+        (response !== null && response.status() === 302) ||
```
lsolcher, line 16: can be response !== null && (response.status() === 301 || response.status() === 302)

app/components/Breadcrumbs.tsx
```diff
@@ -1,88 +1,60 @@
...
-  const location = useLocation();
-
-  const filteredBreadcrumbs = filterBreadcrumbs(breadcrumbs, location.pathname);
-
-  return filteredBreadcrumbs.length > 1 ? (
+    // Otherwise find the route that match the current path, removing trailing slashes
+    return ROUTES.find((r) => r.url === match.pathname.replace(/\/$/, ""));
+  });
+
+  const uniqueBreadcrumbs = breadcrumbs
```
lsolcher, line 23: Are these exact duplicates? We could use Array.from(new Set(breadcrumbs) then

app/components/Breadcrumbs.tsx
HendrikSchmidt: This is the visible Link for the users, no?

app/resources/content/beispiele-regelungen.ts
HendrikSchmidt: The users can also access this page through `beispiele/visualisierungen`, so I think I would leave it as is for now and maybe change it to something like `beispiele/regelungen/regelungsbeispiel` if we ever have a complete page. Wdyt?

app/resources/staticRoutes.ts
```diff
@@ -1,166 +1,187 @@
-const PRE_CHECK_PDF = "digitalcheck-vorpruefung.pdf";
+// These routes can't be exported from ~/routes.ts as that file isn't part of the app environment: https://github.com/remix-run/react-router/issues/12392
+
+import { preCheckQuestions } from "./content/shared/preCheckQuestions";
+
+const PRECHECK_PDF = "digitalcheck-vorpruefung.pdf";
```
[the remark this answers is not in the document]
HendrikSchmidt: You are right, at the moment we don't use them anywhere else. I'll put them in the Route.

app/components/Breadcrumbs.tsx
HendrikSchmidt: Changed :)

app/resources/content/beispiele-regelungen.ts
lsolcher: Sounds good to me

app/components/Breadcrumbs.tsx
HendrikSchmidt: Ah yes, got confused with which Link this is, fixed!

PR #367, opened 2025-04-24
Implement Progressbar

HendrikSchmidt:


app/components/PageHeader.tsx
```diff
@@ -173,7 +189,8 @@ const PageHeader = ({
         >
           {header.items.map((item) => renderDropdownItem(item, "mobile"))}
         </nav>
-        {includeBreadcrumbs && <Breadcrumbs />}
+        {showProgressBar && <ProgressBar />}
+        {includeBreadcrumbs && !hideBreadcrumbs && <Breadcrumbs />}
```
lsolcher, line 193: What about`includeBreadcrumbs = !getFeatureForMatches(matches, "hideBreadcrumbs");` and then `{includeBreadcrumbs  && <Breadcrumbs />}` ?  I find two conditions a bit confusing here

HendrikSchmidt: `includeBreadcrumbs` comes from the props, so we need to check the two conditions at some point. But I could do it outside the jsx:
```
const hideBreadcrumbs = getFeatureForMatches(matches, "hideBreadcrumbs");
const showBreadcrumbs = includeBreadcrumbs && !hideBreadcrumbs;
```

lsolcher (approved):

PR #380, opened 2025-05-13
Footer

bastipnt:
@HendrikSchmidt


app/components/Footer.tsx
```diff
@@ -11,73 +13,95 @@ type LinkProps = {
...
-    </li>
-  );
-
-  const renderLinks = (links: LinkProps[]) => (
-    <ul className="list-unstyled ds-stack ds-stack-16" key={links[0]?.url}>
-      {links.map(renderLink)}
+  className?: string;
+  listClassName?: string;
+}) => (
+  <section className={tailwindMerge("ds-stack ds-stack-8", className)}>
```
bastipnt, line 27: @HendrikSchmidt Why section here? I think div would be the option to use here according to that: https://stackoverflow.com/questions/6939864/what-is-the-difference-between-section-and-div

HendrikSchmidt: I've seen `section` being used in one or two examples for footers, but I agree that a `div` might be a better fit here :)

HendrikSchmidt (commented):
Some minor comments, didn't do a comprehensive review

tests/e2e/general.spec.ts
```diff
@@ -66,6 +70,21 @@ test.describe("test landing page", () => {
   });
 });
 
+test("footer is displayed", async ({ page }) => {
+  await page.goto(ROUTE_LANDING.url);
+  const footerEl = page.getByRole("contentinfo", { name: footer.navLabel });
+  await expect(footerEl).toBeVisible();
+  await expect(
+    footerEl.getByRole("navigation", { name: footer.top.navLabel }),
```
HendrikSchmidt, line 78: This is something I'm always a bit torn on. I lean towards explicitly not importing the content file and repeating the strings to not rely on a specific content file structure here and as a separation between implementation and result. Wdyt?

bastipnt: I totally get this and partially agree.
I just thought about also non developers being able to change wordings without the need to touch the tests?
But I guess it would also be possible to just use a different selector?

HendrikSchmidt (changes requested):
Some small comments, looking good otherwise!

 The separator still merges with the content in some cases. There is also a missing space before "oder":
image

One thing that it's missing is a section in the Readme detailing the vitest snapshot testing. Might also warrant an ADR for the new testing approach.

tests/e2e/general.spec.ts
HendrikSchmidt: I think a different selector would be the best option! Otherwise we've dealt with failing tests as engineers in the past, which of course isn't an optimal solution, but guards against deleted strings, typos etc. No strong feelings here.

app/components/Footer.tsx
```diff
@@ -34,52 +27,93 @@ export default function Footer({ links }: Readonly<FooterProps>) {
...
-                {
-                  preText: "Im Auftrag des",
-                  text: "Bundesministerium des Innern und für Heimat",
-                  url: "https://www.bmi.bund.de/",
-                  openInNewTab: true,
-                },
-              ])}
-            </div>
-          </div>
+      <div className="ds-stack ds-stack-32 sm:ds-stack-40 w-full max-w-[1120px] border-t-2 border-blue-300 py-40">
```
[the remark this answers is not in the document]
bastipnt: Because in the design Sabrina wanted to have a max width of 1120px for the Footer. And there is no predefined with for that.

app/components/Footer.tsx
```diff
@@ -11,75 +12,106 @@ type LinkProps = {
...
+        <nav
+          className="grid grid-cols-1 justify-between gap-y-32 sm:grid-cols-[repeat(3,_minmax(0,_18rem))] sm:grid-rows-2"
+          aria-label={footer.top.navLabel}
+          id="quick-overfiew-nav"
```
HendrikSchmidt, line 67: typo

doc/adr/0012-guidelines-for-testing.md
```diff
@@ -25,3 +25,7 @@ We could do everything between requiring draconian testing standards, to having
 ## Consequences
 
 - Some clean-up & refactoring of existing tests
+
+## Update (2025-05-20)
+
+In we updated our testing guidelines [here](./0020-use-test-ids-instead-of-name-selectors.md)
```
HendrikSchmidt, line 31: Typo

HendrikSchmidt (approved):
Will have a look at the ADR later :)

doc/adr/0020-use-test-ids-instead-of-name-selectors.md
```diff
@@ -0,0 +1,43 @@
...
+- They are particularly useful for UI component rendering (to detect unexpected visual/structural changes) and large data structures (like API responses or generated configurations).
+
+## Consequences
+
+### Benefits
+
+- Emphasis on unit tests provides quicker feedback to developers.
+- Using role-based selectors can reduce test fluctuation.
+- Prioritizing ARIA roles promotes building accessible products.
+
```
HendrikSchmidt, line 39: You could also add quicker running tests/pipelines as an upside

doc/adr/0020-use-test-ids-instead-of-name-selectors.md
```diff
@@ -0,0 +1,43 @@
...
+### Drawbacks/Considerations
+
+- Snapshots can become a maintenance overhead if not managed carefully, especially for frequently changing UI or data.
+- When non developers update texts, snapshots or role-based selectors may be updated to fix failing tests.
```
HendrikSchmidt, line 43: Maybe also add
- the added files that we need to check in
- that we might need to overhaul some existing e2e-tests in the future

doc/adr/0020-use-test-ids-instead-of-name-selectors.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Use ARIA Roles for Test Selectors (Over Test IDs)
+
+- When writing tests that interact with the UI, test selectors should primarily target ARIA roles and accessible names.
+- Test IDs should be considered a fallback, used only when semantic roles are not applicable or sufficient.
+- Encourages the development of more accessible applications by aligning tests with how users (and assistive technologies) perceive and interact with the UI.
+
+### Employ Snapshot Testing where it makes sense
+
+- Snapshot testing will be utilized for verifying the output of UI components or complex data structures when appropriate.
```
HendrikSchmidt, line 28: Could you clarify that you're referring to vitest snapshots here to avoid confusion with playwright?

doc/adr/0020-use-test-ids-instead-of-name-selectors.md
```diff
@@ -0,0 +1,43 @@
+# 20. Updated guidelines for testing
+
+## Status
+
+- 2025-05-20: Proposed
```
HendrikSchmidt, line 5: I think we didn't use this state before and only went with accepted directly, but if you prefer to propose ADRs first I am happy to be convinced :D

doc/adr/0020-use-test-ids-instead-of-name-selectors.md
```diff
@@ -0,0 +1,43 @@
+# 20. Updated guidelines for testing
```
HendrikSchmidt, line 1: This could be a bit more descriptive, e.g. "Use more unit and snapshot tests instead of e2e tests" or something along those lines.

PR #390, opened 2025-05-23
Refactor root and meta exports

HendrikSchmidt:
This PR refactors a few things:
- Uses a `layout` export instead of our homegrown `Document` solution and simplifies `App` and `Error` components
- Moves error texts to own content file
- Moves `<meta>` tags into HTML instead of the root export
- Simplifies the construction of the title meta tags objects and returns it, which simply overrides the existing title of the parent component, instead of dealing with merging logic that was necessary beforehand


bastipnt (commented):
Did have a look, and it looks much cleaner, thanks!
Didn't have ha comment in the files itself, but there are some checks failing.

PR #391, opened 2025-05-26
Staging/70 tage

bastipnt:


tests/e2e/methodPages.spec.ts
HendrikSchmidt, line 1: why are some tests `fixme` and some `skip`?

tests/a11y/general.spec.ts
```diff
@@ -10,7 +10,10 @@ import {
 test.describe("basic example a11y test", () => {
   ROUTES.filter(
     (route) =>
-      !route.url.endsWith(".pdf") && !route.url.endsWith(ROUTE_SUPPORT.url),
+      !route.url.endsWith(".pdf") &&
```
HendrikSchmidt, line 13: could also be something like `![".pdf", ".xlsx", ".docx", ROUTE_SUPPORT.url].some(r => r.url.endsWith)`

app/styles.css
```diff
@@ -7,8 +7,16 @@
...
   --font-sans: BundesSansWeb;
+
+  --color-principle-1: #f9ec9e;
+  --color-principle-2: #f5dcd7;
+  --color-principle-3: #dfd6e2;
+  --color-principle-4: #cce7db;
+  --color-principle-5: #cce4f0;
 }
 
+@source inline("bg-principle-{1..5}");
```
HendrikSchmidt, line 18: could you add a comment what this achieves/why it's necessary?

app/routes/download.$fileName.tsx
```diff
@@ -5,13 +5,18 @@ import trackCustomEvent from "~/utils/trackCustomEvent.server";
...
-  const { fileName } = params;
+  let { fileName } = params;
 
   if (!fileName) {
     // eslint-disable-next-line @typescript-eslint/only-throw-error
     throw new Response("Please provide a file name", { status: 400 });
   }
 
+  // NOTE: 70-tage replace pdf with word document
+  if (fileName === "digitalcheck-begleitende-dokumentation.pdf") {
```
HendrikSchmidt, line 16: is this meant for external links? could you add this as a comment then :)

app/resources/content/methode-ablaeufe-aufgaben-erfassen.ts
```diff
@@ -117,16 +117,30 @@ export const methodsTasksProcesses = {
...
+  //     },
+  //   ],
+  // },
   nextStep: {
     label: "So geht es weiter:",
     icon: ArrowCircleRightOutlined,
-    title: "2.3. IT-Systeme erfassen",
-    text: `Nutzen Sie das Fachwissen der Akteurinnen und Akteure, um die verwendete IT-Infrastruktur für die identifizierten Abläufe zu erfassen und zu verstehen.`,
+    title: "Finden Sie konkrete Möglichkeiten der Digitalisierung",
+    text: `Mit den fünf Prinzipien für Digitaltaugliche Gesetzgebung bekommen Sie auch konkrete Hinweise darauf, worauf Sie achten müssen.`,
```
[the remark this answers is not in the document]
bastipnt: Yes, but it is used in this file everywhere like this. And I'm currently changing the file anyways.

app/components/Badge.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const labelStyle = "ds-label-02-reg rounded-md p-4";
+  const principleStyle = principleNumber
+    ? `bg-principle-${principleNumber}`
+    : "";
+  const hintStyle = hint ? "bg-blue-300 text-blue-800" : "";
+
+  return (
+    <span
+      role="mark"
+      className={twJoin(labelStyle, principleStyle, hintStyle, className)}
```
HendrikSchmidt, line 28: you're joining some overlapping styles here (bg), so `twMerge()` would maybe be the better solution?

tests/e2e/methodPages.spec.ts
bastipnt: I'm actually not sure anymore, will use skip everywhere

app/styles.css
bastipnt: yes, exactly, that was one think I wanted to talk about

app/components/InfoBox.tsx
```diff
@@ -1,41 +1,51 @@
...
 }: InfoBoxProps) => {
   return (
-    <div className="ds-stack ds-stack-8 scroll-my-40" id={identifier}>
-      {label && <Heading {...label} />}
-      {heading && <Heading className="max-sm:ds-heading-02-reg" {...heading} />}
-      <ul
-        className={twJoin(
-          "list-unstyled info-box ds-stack",
-          separator ? "ds-stack-32" : "ds-stack-48",
+    <section className="flex flex-col gap-16 sm:flex-row sm:gap-32">
```
HendrikSchmidt, line 24: why does the flex change here when the icon gets hidden anyways?

app/components/Badge.tsx
```diff
@@ -0,0 +1,35 @@
...
+  hint,
+}: Readonly<LabelProps>) {
+  const labelStyle = "ds-label-02-reg rounded-md p-4";
+  const principleStyle = principleNumber
+    ? `bg-principle-${principleNumber}`
+    : "";
+  const hintStyle = hint ? "bg-blue-300 text-blue-800" : "";
+
+  return (
+    <span
```
HendrikSchmidt, line 26: could this also be a `<mark>` element

app/components/InfoBox.tsx
bastipnt: because it was not hidden before, I added it later ^^
will remove the classes

app/components/InfoBox.tsx
```diff
@@ -1,41 +1,51 @@
...
+      {Icon && <Icon className="hidden size-80 fill-blue-500 sm:block" />}
+      <div className="ds-stack ds-stack-8 scroll-my-40" id={identifier}>
+        {label && (
+          <Badge className="self-start" principleNumber={label.principleNumber}>
```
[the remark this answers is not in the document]
bastipnt: It also works without a `principleNumber`, but I noticed something different, that I want to change here

app/styles.css
HendrikSchmidt: let's keep it that way for now, might change to map in the future

app/components/Badge.tsx
HendrikSchmidt: part of sonar cube analyis

PR #406, opened 2025-06-05
Implement new Visualisation page

bastipnt:


app/components/DropdownMenu.tsx
```diff
@@ -38,7 +39,9 @@ export default function DropdownMenu({
 }: Readonly<DropdownProps>) {
   const isMobile = variant === "mobile";
   const elementId = `dropdown-${label}`;
-  const plausibleTrackingClass = `plausible-event-name=Nav+Bar.${plausibleEventName}.Open+Close`;
+  const plausibleTrackingClass = getPlausibleEvent(
+    `Nav+Bar.${plausibleEventName}.Open+Close`,
```
mpanne, line 43: The naming of the events is now spread across different layers of the application (content and components). Can we think of a way to have it only in one layer?

app/components/HighlightBox.tsx
```diff
@@ -0,0 +1,15 @@
+import Background from "./Background";
+import InfoBox, { InfoBoxProps } from "./InfoBox";
+
+export type HighlightBoxProps = InfoBoxProps;
+
+export default function HighlightBox(props: HighlightBoxProps) {
```
mpanne, line 6: Do we really need a single component here? We had the discussion in the past that we prefer to stick to basic building blocks to keep the markup readable and flexible.

app/components/ImageBox.tsx
```diff
@@ -17,28 +33,52 @@ function ImageBox({
...
       )}
-      <ImageZoomable
-        image={imageForZoomable}
-        plausibleEventName={plausibleEventName}
-      />
-      <figcaption className="label-03-reg pt-[24px]">
-        {image.caption}
-      </figcaption>
-      <hr className="my-[16px] w-[120px] border-t-[2px] border-blue-300" />
+      {zoomable && isBigImage ? (
```
mpanne, line 59: Why can the image only be zoomable if it is of size MEDIUM or LARGE? Can't an image also be small (like a thumbnail) and we still want it to be zoomable? Maybe even because it is presented small by default.

app/components/ImageBox.tsx
```diff
@@ -17,28 +33,52 @@ function ImageBox({
...
+        <ImageZoomable
+          image={imageForZoomable}
+          plausibleEventName={plausibleEventName}
+          className="object-cover"
+          square={!!size}
+        />
+      ) : (
+        <Image {...image} className="object-contain" />
+      )}
+      {image.caption && isBigImage && (
```
mpanne, line 69: Same here, why is the decision of a caption implicitly prevented if it's small? Shouldn't we are in charge if we want to show a caption or not no matter of the size? Shouln't that be fixed by design, not by code?

app/routes/_index.tsx
```diff
@@ -26,10 +26,12 @@ export default function Index() {
             items={index.stepByStep.items}
           />
           <ImageBox
+            className="pb-40"
             image={index.stepByStep.processImage.img}
             plausibleEventName={
               index.stepByStep.processImage.plausibleEventName
             }
+            border
```
mpanne, line 34: TIL: In JSX this is a shortcut to set a boolean to true. Thanks!

app/routes/methoden.$subPage.tsx
```diff
@@ -152,12 +151,16 @@ export default function Index() {
...
+            badge={{
+              text: content.nextStep.label,
+              Icon: content.nextStep.icon,
+            }}
             content={{ markdown: content.nextStep.text }}
             buttons={content.nextStep.buttons}
           />
         </Container>
       )}
+      {/* TODO: fix margin: mt-40 mb-48 */}
```
mpanne, line 163: Open TODO?

app/components/LabelWithIcon.tsx
```diff
@@ -1,33 +0,0 @@
-import { twJoin } from "tailwind-merge";
-
-function LabelWithIcon(
```
mpanne, line 3: Good job replacing this with the Badge component. I like it.

app/resources/content/methode-ablaeufe-aufgaben-erfassen.ts
```diff
@@ -1,146 +1,220 @@
...
+  | ContentVisualisation
+  | ContentItemInfoBox
+  | ContentItemHighlightBox
+  | ContentItemCard;
+
+type TabItem = {
+  title: string;
+  content: ContentItem[];
+  plausibleEventName: string;
+};
```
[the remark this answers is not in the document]
mpanne: Additionally, adding more logic to the content file makes it less usable / maintainable for non-devs I would argue. I might be wrong though.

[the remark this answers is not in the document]
mpanne: I guess we need an informed decision about how to proceed and what the actual pain is we need to solve.

app/routes/methoden.ablaeufe-aufgaben-erfassen.tsx
```diff
@@ -0,0 +1,138 @@
...
+            content.filter((item) => item.element === "visualisation"),
+          )
+          .map(({ documentId }) => documentId),
+      },
+    },
+  });
+
+  if ("error" in visualisationsData) {
+    // eslint-disable-next-line @typescript-eslint/only-throw-error
+    throw new Response(visualisationsData.error, { status: 400 });
```
mpanne, line 48: I would argue that this should have status 500 since it is not the users fault but an internal server error.

mpanne: Additionally, I would say we should use another error message here and not forward the strapi error since it is shown to the user and might contain information we do not want to disclose. So the approach I would suggest: log the strapi error response and throw an error like "Hier hat leider etwas nicht funktioniert, das tut uns Leid."

app/routes/methoden.ablaeufe-aufgaben-erfassen.tsx
```diff
@@ -0,0 +1,138 @@
...
+            <Card
+              key={(contentItem.props.heading?.text as string) + i}
+              {...contentItem.props}
+            />
+          );
+        } else if (contentItem.element === "visualisation") {
+          const visualisationsItem = visualisationsData.find(
+            ({ documentId }) => documentId === contentItem.documentId,
+          );
+          if (!visualisationsItem) return <></>;
```
mpanne, line 85: Maybe use a switch statement for better readability?

mpanne (changes requested):
Good work, thank you! 

Biggest topic for me is the logic in the content layer. I feel uncomfortable to violate the separation of concerns here and use representation logic in the content layer. I would love to find a way to keep that clean. I definitely see that this is a bigger issue across the application and it did not start here. I guess we need to re-open the content management discussion soon.

app/components/DropdownMenu.tsx
bastipnt: Yes, I already have the feeling that it has become complex and rather hard to keep track of in the code.
We have a spike next sprint to try out a different tracking solution.
I would keep it open until we have a decision there and then come up with a better approach if we keep Plausible

app/resources/content/methode-ablaeufe-aufgaben-erfassen.ts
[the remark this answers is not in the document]
bastipnt: Yes, I totally agree. How should I go forward with this? Refactor it to not use logic in the resources files?

app/components/DropdownMenu.tsx
mpanne: Yes, agree.

app/components/HighlightBox.tsx
mpanne: Alright. Would you add a tech debt ticket in our backlog to replace all the existing occurrences of that pattern so that we do not mix both approaches?

app/components/ImageBox.tsx
mpanne: Ok. I still do not really see the point to restrict it via code. It's more code and could lead to odd behavior if the design system changes some day. But I am fine with keeping it.

mpanne (approved):
Good job!

PR #459, opened 2025-07-18
Remove background component

bastipnt:


app/components/Card.tsx
```diff
@@ -8,20 +9,24 @@ export type CardProps = BoxProps & {
...
-        <div className="px-20 pt-40 sm:px-96 sm:pt-44">
-          <div className="&_img:object-cover &_img:object-top h-0 overflow-hidden pb-[40%] shadow-2xl">
-            <Image {...image} />
-          </div>
-        </div>
-      </Background>
-      <Background
-        backgroundColor="blue"
-        className="px-16 py-24 sm:px-80 sm:pt-40 sm:pb-48"
+    <div className="overflow-hidden rounded-lg">
```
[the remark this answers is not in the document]
bastipnt: Ah yes, I saw that the definition of the `Card` in the design system has rounded boarders, that's why I fixed it.

app/routes/hinweis.tsx
```diff
@@ -1,17 +1,16 @@
-import Background from "~/components/Background";
 import Container from "~/components/Container";
 import InfoBoxList from "~/components/InfoBoxList";
 import { hinweis } from "~/resources/content/hinweis";
 
 export default function Index() {
   return (
-    <Background backgroundColor="blue" className="parent-bg-blue">
+    <div className="parent-bg-blue">
```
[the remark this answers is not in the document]
bastipnt: yess exaclty

mpanne (approved):

app/components/Container.tsx
```diff
@@ -1,47 +1,23 @@
...
-    className,
-  );
+  let cssClasses = twMerge("container pt-40 pb-48", className);
 
-  if (backgroundColor === "default") {
-    return <div className={cssClasses}>{children}</div>;
-  }
-
-  if (backgroundColor && overhangingBackground) {
+  if (overhangingBackground) {
```
HendrikSchmidt, line 16: This could be incorporated into the `cssClasses` at the beginning, no?

```
const cssClasses = twMerge("container pt-40 pb-48", overhangingBackground && "rounded-lg container-overhanging-background", className);
```

app/components/FeedbackForm.tsx
```diff
@@ -157,23 +156,20 @@ export default function FeedbackForm(props: Readonly<FeedbackFormProps>) {
...
     );
   }
 
   return (
-    <Background
-      backgroundColor="blue"
-      className={twJoin("pt-40 pb-48", props.className)}
-    >
-      <Container backgroundColor="white" overhangingBackground>
+    <div className={twJoin("bg-ds-blue pt-40 pb-48", props.className)}>
```
HendrikSchmidt, line 171: Not your job with this PR, but maybe a `twMerge` would be better to allow more customizability?

app/components/Hero.tsx
```diff
@@ -1,6 +1,5 @@
 import { ReactNode } from "react";
-import { BackgroundColor } from ".";
-import Background from "./Background";
+import customTwMerge from "~/utils/tailwindMerge";
```
HendrikSchmidt, line 2: Also not your job, but using `twMerge` instead of `customTwMerge` everywhere would be more consistent.

app/components/InfoBox.tsx
```diff
@@ -42,13 +43,17 @@ const InfoBox = ({
...
   Icon,
+  look,
 }: InfoBoxProps) => {
   return (
     <div
       id={identifier}
       data-testid="info-box-container"
       className={twMerge(
         "flex scroll-my-40 flex-col gap-32 sm:flex-row",
+        look === "highlight" && `bg-ds-blue rounded-lg px-16 py-40 sm:px-80`,
```
HendrikSchmidt, line 54: I think quotes instead of backticks would be more suitable here, I never understand why the linter doesn't fix that.

app/routes/unterstuetzung/route.tsx
```diff
@@ -232,9 +230,9 @@ export default function Index() {
             <ButtonContainer buttons={supportHow.buttons} />
           </div>
         </Container>
-      </Background>
+      </div>
       <div id="angebote">
-        <Background backgroundColor="white">
+        <div className="bg-white">
```
HendrikSchmidt, line 235: Could be merged with the parent div maybe?

app/styles.css
```diff
@@ -14,9 +14,14 @@
   --color-principle-5: #cce7db; /* order 4 */
   --color-principle-3: #cce4f0; /* order 5 */
 
-  --color-principle-4: var(
-    --color-orange-500
-  ); /* order 6 -> not used anymore */
+  /* Dito default colors: */
```
HendrikSchmidt, line 17: We could use `--color-*: initial` to completely disable all of the default colors and force developers to first assign a color a name before using one offs. Wdyt? Would need a check if we use different colors for fonts etc too ofc.

app/components/ListItem.tsx
```diff
@@ -93,42 +92,34 @@ const ListItem = ({
...
                 )}
                 role="none"
               ></div>
             )}
           </div>
-          <div className={twJoin(backgroundColor && "w-full rounded-lg")}>
-            <Background
-              backgroundColor={(backgroundColor as BackgroundColor) || "white"}
+          <div className={twJoin(backgroundColorClass && "w-full rounded-lg")}>
+            <div
```
HendrikSchmidt, line 102: Can these two divs be merged?

HendrikSchmidt (changes requested):
Thanks for the improvements, I like it a lot better! I've added a few small comments, some of them not really your responsibility, so feel free to delegate them to a ticket instead. I also ran a quick search to evaluate which kind of `bg-` classes we're still using and apart from the "official" ones, there are still occurrences of:
```
bg-[#282828]
bg-[#B3C9D6]
bg-[#F9E5EC]
bg-[0%_35%]
bg-[rgba(255,255,255,0.50)]
bg-blue
bg-blue-100
bg-blue-200
bg-blue-300
bg-blue-400
bg-blue-500
bg-blue-800
bg-blue-900
bg-gray-100
bg-green-100
bg-principle-4
bg-white/70
bg-yellow-200
bg-yellow-300
```
I think it could make sense to either map them to the `bg-ds-` classes or create new classes for them.

mpanne (approved):
Lovely! I like it very much with using the tailwind theme. Only some small remarks.

app/styles.css
```diff
@@ -14,9 +14,14 @@
...
+  --color-ds-blue: #f2f6f8; /* blue-100 */
```
HendrikSchmidt, line 18: I find the naming scale a bit confusing now. Maybe `ds-light-blue` would be more suitable here?

app/components/Container.tsx
```diff
@@ -1,47 +1,23 @@
...
 import twMerge from "~/utils/tailwindMerge";
-import { BACKGROUND_COLORS, type BackgroundColor } from ".";
 
 type ContainerProps = {
-  backgroundColor?: BackgroundColor;
   overhangingBackground?: boolean;
   className?: string;
 };
 
 export default function Container({
```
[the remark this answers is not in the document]
HendrikSchmidt: I had a similar thought, maybe something to discuss in the engineering checkin.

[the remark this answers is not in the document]
mpanne: I added it to the agenda.

app/components/Container.tsx
bastipnt: yess, done

HendrikSchmidt (approved):
lgtm

PR #484, opened 2025-07-31
Add principle poster box

HendrikSchmidt:


app/components/InfoBox.tsx
```diff
@@ -42,7 +55,7 @@ const InfoBox = ({
   detailsSummary,
   linkList,
   buttons,
-  Icon,
+  visual,
```
bastipnt, line 58: nice, I think I like this naming

app/components/InfoBox.tsx
```diff
@@ -28,11 +26,26 @@ export type InfoBoxProps = {
...
+  | { type: "icon"; Icon: React.FC<React.SVGProps<SVGSVGElement>> }
+  | { type: "image"; image: InfoImageProps }
+  | { type: "none" };
+
+export type InfoBoxProps = BaseInfoBoxProps & {
+  visual?: VisualProps;
+};
+
+const imageSizes = {
+  icon: "size-80",
```
bastipnt, line 43: makes sense to call it icon and not xs 👍

app/components/PrinciplePosterBox.tsx
```diff
@@ -0,0 +1,41 @@
...
+          tagName: "h2",
+          text: principlePosterBox.heading,
+        }}
+        content={principlePosterBox.content}
+        linkList={{
+          links: [
+            {
+              title: principlePosterBox.downloadTitle,
+              url: principlePosterBox.downloadUrl,
+              download: true,
```
bastipnt, line 34: Linter is complaining, that `download` does not exist in the props 🤷

.talismanrc
```diff
@@ -4,4 +4,6 @@ fileignoreconfig:
 - filename: .env.test
   ignore_detectors: [filename]
 - filename: app/root.tsx
-  allowed_patterns: [posthogKey]
\ No newline at end of file
+  allowed_patterns: [posthogKey]
```
bastipnt, line 7: does this belong to this pr?

bastipnt (changes requested):
It already looks really good, just two things:
1. The `Image` should move to the bottom of the `InfoBox` on mobile, according to figma: https://www.figma.com/design/HiiKWvM5xBlNCM9wXYvInb/%F0%9F%9F%A6-Digitalcheck-Komponenten-Library?node-id=1334-7921
2. The `Image` on `InfoBox` is not fully responsive: Image
Apart from that nice work :)

.talismanrc
HendrikSchmidt: I think this is only highlighted as added because I put a newline at the end :)

HendrikSchmidt:
> It already looks really good, just two things:
> 
> 1. The `Image` should move to the bottom of the `InfoBox` on mobile, according to figma: https://www.figma.com/design/HiiKWvM5xBlNCM9wXYvInb/%F0%9F%9F%A6-Digitalcheck-Komponenten-Library?node-id=1334-7921
> 2. The `Image` on `InfoBox` is not fully responsive

1. Changed the `flex-direction` to achieve this.
2. Added responsive image sizing for large images.

PR #488, opened 2025-08-04
Staging/beispielseiten

bastipnt:


app/components/ParagraphList.tsx
```diff
@@ -165,180 +128,32 @@ function Paragraph({ paragraph }: Readonly<{ paragraph: Paragraph }>) {
...
-  const getAbsatzGroupTitle = (absatzGroup: AbsatzWithNumber[]) =>
-    absatzGroup.length > 1
-      ? `(${absatzGroup[0].number}) – (${absatzGroup[absatzGroup.length - 1].number})`
-      : `(${absatzGroup[0].number})`;
-
   // renders the interlaced Absaetze with relevant PrinzipErfuellungen and grouped Absaetze without relevant PrinzipErfuellungen
   return (
     <div key={paragraph.Nummer}>
-      <div className="ds-stack ds-stack-8">
+      <div className="flex flex-col gap-24 md:gap-32">
```
HendrikSchmidt, line 134: Could be merged with parent?

app/components/ParagraphList.tsx
```diff
@@ -165,180 +128,32 @@ function Paragraph({ paragraph }: Readonly<{ paragraph: Paragraph }>) {
...
-                }
-              />
-            ) : (
-              <Absatz key={absatzGroup.id} absatz={absatzGroup} />
-            ),
-          )}
+        {paragraph.Titel && (
+          <p className="ds-subhead font-bold">{paragraph.Titel}</p>
+        )}
+        <div className="flex flex-col gap-24 md:gap-32">
```
HendrikSchmidt, line 144: Maybe use `space-y-...` instead of flex + gap here?

app/components/PrincipleHighlightModifier.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+    return () => removeEventListener("resize", handleResize);
+  }, []);
+
+  const [text, numberGroup] = node.text ? node.text.split(/(\[\d])/g) : [];
+  if (!numberGroup || !explanationIdPrefix) return text;
+
+  const principleNumber = Number(
+    numberGroup[1],
+  ) as keyof typeof HIGHLIGHT_COLORS;
```
HendrikSchmidt, line 32: `HIGHLIGHT_COLORS` seems a bit counterintuitive here.

app/components/Absatz.tsx
```diff
@@ -0,0 +1,164 @@
...
+import PrincipleHighlightModifier from "./PrincipleHighlightModifier";
+
+type AbsatzProps = {
+  absatz: AbsatzWithNumber | AbsatzWithNumber[];
+  principlesToShow: Prinzip[];
+  useAnchorLinks: boolean;
+};
+
+export default function Absatz({
+  absatz,
```
HendrikSchmidt, line 29: maybe `absatzOrAbsatzGroup` could provide clarity here

app/components/Absatz.tsx
```diff
@@ -0,0 +1,164 @@
...
+      </p>
+    </div>
+  );
+}
+
+type AbsatzWithoutErfüllungenProps = {
+  absatzGroup: AbsatzWithNumber[];
+};
+
+function AbsatzWithoutErfüllungen({
```
HendrikSchmidt, line 135: this always renders a group (sometimes just containing 1), so maybe a different name could make that more explicit

app/components/Absatz.tsx
```diff
@@ -0,0 +1,164 @@
...
+  principlesToShow: Prinzip[];
+  useAnchorLinks: boolean;
+};
+
+export default function Absatz({
+  absatz,
+  principlesToShow,
+  useAnchorLinks,
+}: Readonly<AbsatzProps>) {
+  const isAbsatzWithoutErfuellungen = Array.isArray(absatz);
```
HendrikSchmidt, line 33: `isAbsatzGroup...`

app/components/Absatz.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const content = prependNumberToAbsatz(absatz);
+
+  return (
+    <PrincipleHighlightProvider
+      absatzId={absatz.id}
+      principlesToShow={principlesToShow}
+      useAnchorLinks={useAnchorLinks}
+    >
+      <div className="flex flex-col gap-16 md:grid md:grid-cols-[5fr_3fr] md:gap-32 [&_ol]:list-decimal [&_ol_ol]:list-[lower-alpha]">
+        <div className="italic">
```
HendrikSchmidt, line 47: maybe give `BlocksRenderer` `className` props instead?

app/components/Absatz.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const explanationClasses = twJoin(
+    "w-fit max-w-[642px] px-8",
+    color.border,
+    shouldHighlight ? "border-4 p-4" : "border-l-4 p-8",
+  );
+
+  return (
+    <div className={explanationClasses} id={id} data-testid={id}>
+      <div className="flex content-center gap-8">
+        <div>
```
HendrikSchmidt, line 107: why is this needed?

app/components/ParagraphList.tsx
HendrikSchmidt, line 1: The architecture still needs updating

HendrikSchmidt (commented):
The changes seem sensible to me as a preparation to render individual Absätze.

I left some smaller comments. Also, as a side note for design, I think we could make more use of the horizontal space.

Image

app/components/Absatz.tsx
bastipnt: Because the BlocksRenderer itself does not wrap its children. Without a wrapper the flex would be applied to all elements, but we do not want that.

app/components/PrincipleHighlightModifier.tsx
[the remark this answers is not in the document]
bastipnt: I have it separated, because I want to use it for the Principle page, for Absaetze, that will not show an explanation. This will probably be a different component which also uses the PrincipleHighlightModifier

app/components/PrincipleHighlightModifier.tsx
```diff
@@ -0,0 +1,79 @@
...
+    );
+
+  return (
+    <Link
+      replace
+      to={`#${explanationID(explanationIdPrefix, principleNumber)}`}
+      onClick={() => {
+        setActiveHighlight({ id: highlightID, principleNumber });
+      }}
+      aria-label={`Erfüllt Prinzip ${principleName}: ${text}`}
```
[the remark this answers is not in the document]
bastipnt: where would you put it then?

app/components/ParagraphList.tsx
bastipnt: done

app/components/PrincipleHighlightModifier.tsx
HendrikSchmidt: I get that, but I think this doesn't actually provide type safety, right? How about something like this:
maybe in `utils/strapiData.server`
```TS
type PrincipleNumber = Prinzip["Nummer"];

const isValidPrincipleNumber = (num: number): num is PrincipleNumber =>
  num >= 1 && num <= 5 && Number.isInteger(num);
```

in `PrincipleHighlightModifier.tsx`
```TS
if (!isValidPrincipleNumber(principleNumber)) return text;
```
 
We could also then use the `type PrincipleNumber` for indexing `HIGHLIGHT_COLORS`. Also, currently the `Prinzip["Nummer"]` is only 1-5 and needs to add the 6 and remove the 4 (I think).

app/components/Absatz.tsx
HendrikSchmidt: see comment about wrapper, if possible I think that would be cleaner

app/components/PrincipleHighlightModifier.tsx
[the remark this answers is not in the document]
HendrikSchmidt: Sorry, let me rephrase: I think the link should in fact not get an `aria-label`, instead the `mark` (in all cases) should get [aria-details](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-details) that links to the id of the explanation on the right as reference.

app/components/ParagraphList.tsx
```diff
@@ -37,21 +21,13 @@ import { BlocksRenderer } from "./BlocksRenderer";
...
- │    ├── Absatz
- │    │    │ Renders an Absatz with relevant principles, highlights and matching explanations.
- │    │    ├─ Context provided: explanationIdPrefix
- │    │    ├─ Processing: Adds the number of the Absatz to the text.
- │    │    │
+ │    ├── AbsatzOrAbsatzGroup
+ │    │    │ Renders an Absatz or an AbsatzGroup with relevant principles, highlights and matching explanations.
+ │    │    ├─ Context handled by PrincipleHighlightProvider: principlesToShow, activeHighlight, setActiveHighlight, explanationIdPrefix, useAnchorLinks
+ │    │    ├─ Processing: Differs between Absetze with relevant principles and groups of Absaetze without.
  │    │    ├── BlocksRenderer (with modifiers: { underline: PrincipleHighlight })
```
HendrikSchmidt, line 31: Should read `PrincipleHighlightModifier`, no?

app/components/ParagraphList.tsx
HendrikSchmidt: I think this doesn't accurately represent the structure anymore, especially for how Absätze are rendered, due to the new wrapping where we either have a `BlocksRenderer` + 1..n `PrincipleExplanation` or a `AbsatzGroupWithoutErfüllungen`. Do you know what I mean?

I have a few nitpicks, lmk if they're too annoying, then I can fix them too:
- I prefer having an extra line above each new component to make it clearer where a new component starts (e.g. over Paragraph and AbsatzOrAbsatzGroup)

app/routes/grundlagen.fuenf-prinzipien.tsx
```diff
@@ -51,7 +51,7 @@ export default function FundamentalsFivePrinciples() {
             }}
             badge={{
               children: principle.label,
-              principleNumber: principle.principleNumber as PrincipleNumber,
+              principleNumber: principle.principleNumber as PrinzipNummer,
```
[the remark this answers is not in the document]
bastipnt: same here (will change)

app/providers/PrincipleHighlightProvider.tsx
```diff
@@ -0,0 +1,56 @@
+import { createContext, ReactNode, useState } from "react";
+import { Prinzip } from "~/utils/strapiData.server";
+import { Optional } from "~/utils/utilityTypes";
+
+export type PartialPrinzip = Optional<
```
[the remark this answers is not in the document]
bastipnt: yes, I'm already doing this in the next Step. Can we leaf it like this for now, until I also updated the method pages?

app/components/PrincipleHighlightModifier.tsx
```diff
@@ -0,0 +1,89 @@
...
+}: Readonly<PrincipleHighlightProps>) {
+  return (
+    <mark
+      id={id}
+      className={twJoin(
+        "ds-body-01-reg",
+        PRINCIPLE_COLORS[principle.Nummer].background,
+      )}
+      aria-label={`Textbeispiel erfüllt Prinzip: ${principle.Name}`}
+      aria-details={absatzId && explanationID(absatzId, principle.Nummer)}
```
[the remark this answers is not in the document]
bastipnt: yes

PR #1013, opened 2026-08-03
feat(style): switch from Angie Design System to KERN

ehrengretl:


ehrengretl:
@ChaosCoder Hier der PR.

Viele Änderungen durch die Farbklassen.
Aber hier eine kleine Zusammenfassung: Viel passiert ist bei den Button, den Form Input Elementen & dem Footer. Zusätzlich zwei neue Dateien kern-utils.css & fonts.css.  

Zum mergen: ich würde es tatsächlich in einen commit squashen

src/components/InputError.tsx
```diff
@@ -9,23 +9,43 @@ type InputErrorProps = PropsWithChildren<{
...
+    //     look === "warning" && "bg-yellow-300 inline-flex px-4 text-black",
+    //   )}
+    // >
+    // {look === "error" && (
+    //   <div className="shrink-0">
+    //     <ErrorOutline className="fill-ds-error" />
+    //   </div>
+    // )}
+    //   <span className="sr-only">Fehler:</span> {children}
+    // </div>
```
ChaosCoder: Is there a reason you left this in?

src/components/InfoBox.tsx
```diff
@@ -121,12 +121,12 @@ const LinkList = ({ links }: { links: ContentLink[] }) => (
...
-          <DownloadLinkButton key={to} href={to} omitIcon {...rest}>
-            {text}
-          </DownloadLinkButton>
-        );
+      // if (link.download)
+      //   return (
+      //     <DownloadLinkButton key={to} href={to} {...rest}>
+      //       {text}
+      //     </DownloadLinkButton>
+      //   );
```
ChaosCoder: Should we delete this?

src/components/InputError.tsx
```diff
@@ -9,23 +9,43 @@ type InputErrorProps = PropsWithChildren<{
...
+          <span className="kern-body">{children}</span>
+        </p>
       )}
-    >
       {look === "error" && (
         <div className="shrink-0">
           <ErrorOutline className="fill-ds-error" />
         </div>
       )}
-      <span className="sr-only">Fehler:</span> {children}
```
ChaosCoder, line 27: Are we still showing the actual error somewhere? We did before it seems but now it's gone?

Maybe we could introduce a test on Input/Combobox that asserts that we show the error? That would have caught it.

src/components/Button.tsx
```diff
@@ -162,36 +145,27 @@ export function LinkButton({
...
-    <SaveAltOutlinedIcon className="fill-current" />
-  );
+  href,
+  children,
+  className,
+}: Readonly<DownloadLinkButtonProps>) {
   return (
-    <LinkButton
-      iconLeft={iconLeft}
-      download
```
ChaosCoder, line 192: We are missing the `download` attribute here, I guess that's not intended, right?

src/components/Button.tsx
```diff
@@ -89,16 +79,14 @@ function Button({
...
-      {...props}
-      type={type}
-      id={id}
-      data-testid={id}
-      className={buttonClasses}
-    >
-      {formatIcon(iconLeft)}
-      <span className="ds-button-label">{children}</span>
-      {formatIcon(iconRight)}
+    <button {...props} id={id} data-testid={id} className={buttonClasses}>
```
ChaosCoder: Sonar already mentions it, a button must have a type defined. Why did you get rid of it and can we assume the type here or do we need to pass it?

src/resources/content/vorpruefung.ts
```diff
@@ -162,10 +162,10 @@ export const preCheck = {
         headline: "Warum digitaltaugliche Regelungen schreiben?",
         content: dedent`
           <p><strong>Digitaltaugliche Regelungen sparen Ressourcen</strong></p>
-          Eine gute digitale Umsetzung spart langfristig Zeit und sorgt dafür, dass Ziel und Wirkung des Vorhabens erreicht werden - aufseiten der Normadressatinnen und Normadressaten und aufseiten der Verwaltung.
+          <p>Eine gute digitale Umsetzung spart langfristig Zeit und sorgt dafür, dass Ziel und Wirkung des Vorhabens erreicht werden - aufseiten der Normadressatinnen und Normadressaten und aufseiten der Verwaltung.</p
```
ChaosCoder:
```suggestion
          <p>Eine gute digitale Umsetzung spart langfristig Zeit und sorgt dafür, dass Ziel und Wirkung des Vorhabens erreicht werden - aufseiten der Normadressatinnen und Normadressaten und aufseiten der Verwaltung.</p>
```

src/components/RadioGroup.tsx
```diff
@@ -67,33 +66,35 @@ function RadioGroup<FormData, Value extends string | number = string>({
                 "aria-invalid": hasError || hasWarning,
                 "aria-required": required,
                 className: twMerge(
-                  "ds-radio self-start",
-                  hasError && "has-error",
-                  hasWarning && "has-warning",
+                  "kern-form-check__radio self-start",
+                  (hasError || hasWarning) &&
```
ChaosCoder, line 71: That means we render warnings the same as errors (in yellow). Is that intended from the design?

src/components/Textarea.tsx
```diff
@@ -47,21 +46,26 @@ function Textarea({
...
+        <p className="kern-error" id={errorId}>
+          <span
+            className={twMerge(
+              "kern-icon kern-icon--warning kern-icon--md",
+              (hasWarning || hasError) && "!bg-yellow-700",
+            )}
+            aria-hidden="true"
+          ></span>
+          <span className="kern-body">{error || field.error()}</span>
+        </p>
```
ChaosCoder: Same as the other comment about same styling of error and warning. Intended? It's hope the user never sees errors but I guess it would be good to differentiate?

ChaosCoder (commented):
Overall looks nice and aside the small comments mergeable. Feel free to address the issues and then merge without a further review from me :)

src/components/Button.tsx
ehrengretl: We had multiple Button sub-classes here. The Download Button is now visualized with a KERN tertiary Button and an icon.

src/components/Textarea.tsx
ehrengretl: 1. fix: refactored the components to also use the InputError Component.
2. fix: InputError component also renders the the error message in case of an error.
The usage of Error and warnings usage within the Form components is a bit confusing. Almost every component uses the warningInsteadOfError flag, which simply treats all errors as warnings.
Also the error/warning validation is a bit buggy. However, I would suggest to go over these architecture once we start refactoring the Digitalcheck?

Styling-wise, errors are intended to be displayed in yellow, so the underlying logic will eventually need to be refactored.

